PR #13, opened 2019-06-06
Attempt to find `kubetools.yml` in all directories backup to root.

### LINKED ISSUE #11: Work in subdirectories
ie like a git repo, so if inside an `app/blah`, it'll pick up `app/kubetools.yml`.

Fizzadar:
Only happens when directory is not explicitly set, ie in dev.

Closes: #11

### DIFF
--- a/kubetools_client/config.py
+++ b/kubetools_client/config.py
@@ -3,7 +3,7 @@ Shared Python module for handling Kubetools (kubetools.yml) config. Used by both
 the dev client (ktd) and the server (lists KubetoolsClient as a requirement).
 '''
 
-from os import path
+from os import getcwd, path
 
 import six
 import yaml
@@ -52,10 +52,34 @@ def load_kubetools_config(
         dev (bool): filter config items by dev mode
     '''
 
-    possible_files = (
+    possible_filenames = (
         'kubetools.yml',
         'kubetools.yaml',
     )
+
+    if directory:
+        possible_files = [
+            path.join(directory, filename)
+            for filename in possible_filenames
+        ]
+    else:
+        # Look in this directory
+        directory = getcwd()
+        possible_files = [
+            path.join(directory, filename)
+            for filename in possible_filenames
+        ]
+
+        # Attempt parent directories back up to root
+        while True:
+            directory, splitdir = path.split(directory)
+            if not splitdir:
+                break
+
+            possible_files.extend([
+                path.join(directory, filename)
+                for filename in possible_filenames
+            ])
 
     config = None
 
--- a/kubetools_client/config.py
+++ b/kubetools_client/config.py
@@ -63,7 +63,6 @@ def load_kubetools_config(
             for filename in possible_filenames
         ]
     else:
-        # Look in this directory
         directory = getcwd()
         possible_files = []
 
--- a/kubetools_client/config.py
+++ b/kubetools_client/config.py
@@ -65,29 +65,23 @@ def load_kubetools_config(
     else:
         # Look in this directory
         directory = getcwd()
-        possible_files = [
-            path.join(directory, filename)
-            for filename in possible_filenames
-        ]
+        possible_files = []
 
         # Attempt parent directories back up to root
         while True:
-            directory, splitdir = path.split(directory)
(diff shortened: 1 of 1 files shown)


kubetools_client/config.py
```diff
@@ -52,11 +52,35 @@ def load_kubetools_config(
...
 
-    possible_files = (
+    possible_filenames = (
         'kubetools.yml',
         'kubetools.yaml',
     )
 
+    if directory:
+        possible_files = [
+            path.join(directory, filename)
```
chrissinclairedited, line 62: Hmm looks like this logic is now duplicated on L89?

kubetools_client/config.py
```diff
@@ -52,11 +52,35 @@ def load_kubetools_config(
...
+    else:
+        # Look in this directory
+        directory = getcwd()
+        possible_files = [
+            path.join(directory, filename)
+            for filename in possible_filenames
+        ]
+
+        # Attempt parent directories back up to root
+        while True:
```
chrissinclairedited, line 70: Hmmm only thought here is that this will continue to read outside of any repo. I wonder if we only want to read up to a certain point? E.g. my repos are typically laid out as
```
~
    - source
        - repos
            - project_1
                - some_sub_dir
            - project_2
                - some_other_sub_dir
```

Suppose I'm in `~/source/repos/project_1/some_sub_dir/`, then it feels like we should only check as far as `~/source/repos/project_1`, and not look for `~/kubetools.yml`, `~/source/kubetools.yml` or `~/source/repos/kubetools.yml`. Not sure how best to implement this - we could do `git rev-parse --show-toplevel`, but that assumes being in a git repository. Or maybe it's not an issue and we shouldn't worry. Thoughts?

chrissinclairedited (changes requested):
Thanks @Fizzadar - there's one small change needed and one question I have for you, but thanks for this, will be so handy!

kubetools_client/config.py
gchazot: Agreed, but I can't see any clean solution to the problem here

kubetools_client/config.py
gchazot: Well, this is now also in line 69 and 80.
Looks to me like a perfect situation for a good old recursive call

kubetools_client/config.py
Fizzadar: Yeah this is really hard/impossible to solve I think - can't see any way of knowing the correct point to stop looking "up" a directory. Is this better or worse than the current directory only lookup?

chrissinclairedited: I'd still say this is better - I find myself far more often wanting to use `ktd` from within a subdirectory than creating `kubetools.yml` outside a repo, and will happily allow it to check for some files that don't exist if it allows that functionality.

Really the only case I can see this being an issue is with git submodules (where both repos have a `kubetools.yml`), but that's a real edge case and if we're doing that we should expect a few gotchas like this.

kubetools_client/config.py
Fizzadar: This should be implemented in https://github.com/EDITD/kubetools-client/pull/13/commits/dd76be1b4f424bbcf1ad2d050d6fa7009967cd0e!

Fizzadar:
@chrissinclairedited @gchazot thanks for the reviews - I've tidied up the possible files list generation!

chrissinclairedited (approved):
Thanks for this @Fizzadar - should make life a load better! LGTM!

kubetools_client/config.py
```diff
@@ -65,29 +65,23 @@ def load_kubetools_config(
     else:
         # Look in this directory
```
gchazot: This comment is now obsolete

gchazot (approved):

kubetools_client/config.py
Fizzadar: https://github.com/EDITD/kubetools-client/pull/13/commits/dbc73ee5efef9eba7a9a41b4206a18b828395dc7

Fizzadar:
Thanks @chrissinclairedited @gchazot!

PR #21, opened 2019-10-07
Split out docker-compose specific code into "dev backend".

### LINKED ISSUE #23: Abstract docker-compose into "dev backends"
This will enable using different backends (docker-compose/k3s/minikube) for local development.

Fizzadar:
Long run this opens up the possibility of using a local K3S cluster, minikube
or even a kull K8S install as a dev backend.

For now it just moves things about!

Closes: #23


kubetools_client/dev/backends/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+DEV_BACKEND_PROVIDERS = {
+    'docker_compose': docker_compose,
+}
+
+
+# Now, using settings splice in the .backend module to match the provider
+settings = get_settings()
+
+try:
+    backend = DEV_BACKEND_PROVIDERS[settings.DEV_BACKEND]
```
chrissinclairedited, line 17: I'm guessing long run this will pull in from `kubetools.yml`?

chrissinclairedited (approved):
LGTM! Thanks for this @Fizzadar

kubetools_client/dev/backends/__init__.py
Fizzadar: Unsure yet! So thinking it's more "machine specific" so would be part of `~/.kubetools/kubetools.conf` rather than project specific. 

Ideally this would mean two engineers could work on the same project - one using docker compose as is today and another using k3s; all without any specific changes depending on which one.

chrissinclairedited: Ah I see! Is there something driving that or just as a nice to have kind of feature?

gchazot (approved):
LGTM

kubetools_client/dev/backends/__init__.py
Fizzadar: Very much nice-to-have! Saw a thing about local dev with K3S and seems sensible to be flexible w.r.t this!

Fizzadar:
Thanks for the reviews @chrissinclairedited @gchazot!

PR #22, opened 2019-10-07
Move Kubernetes related config generation from server -> client.

### LINKED ISSUE #18: Move Kubernetes config generation into the client
Currently implemented @ https://github.com/EDITD/kubetools/tree/develop/kubetools/kubebuilder/util/config

Would be neat if the client could generate these config files - reduces the complexity of the server and means one could do everything w/o the server.

Fizzadar:
Moves over the logic that turns a `kubetools.yml` config into Kubernetes objects.

Keeps all our "config" logic in the client (compose vs k8s). Also means we can use these to generate dev configs when/if we implement a k[38]s dev backend (https://github.com/EDITD/kubetools-client/pull/21).

Closes: #18


Fizzadar:
Going to merge this as it's really just moving code from server -> client.

PR #24, opened 2019-10-07
Fix issue where dev exceptions would not be formatted correctly

chrissinclairedited:
Because the `except KubeError` was first, and `KubeDevCommandError` and `KubeDevError` are both `KubeError`'s, we were never using those `except` blocks. By moving this later, we use the proper `KubeDevCommandError` handler which correctly formats `stdout` (as it's being passed in raw, rather than through the `__str__` handler for `KubeError`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # WIP/Unreleased
 
+- Fix issue where stdout from a kubetools dev exception would not be formatted properly
+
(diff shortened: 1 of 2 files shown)


Fizzadar (approved):
Nice catch @chrissinclairedited! LGTM

chrissinclairedited:
Thanks for the review @Fizzadar!

PR #47, opened 2020-07-13
Add file and ignore-git-commit flags

stroebs:
Add two flags to deploy command:
- `-f` / `--file`: Input a custom `kubetools.yml` file path to deploy from
- `--ignore-git-changes`: Ignore uncommitted changes to git when deploying

### DIFF
--- a/kubetools/config.py
+++ b/kubetools/config.py
@@ -33,6 +33,7 @@ def load_kubetools_config(
     env=None,
     namespace=None,
     dev=False,  # when true disables env/namespace filtering (dev *only*)
+    custom_config_file=False,
 ):
     '''
     Load Kubetools config files.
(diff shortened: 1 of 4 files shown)


kubetools/config.py
```diff
@@ -50,10 +51,14 @@ def load_kubetools_config(
         dev (bool): filter config items by dev mode
     '''
 
-    possible_filenames = (
-        'kubetools.yml',
-        'kubetools.yaml',
-    )
+    if custom_config_file:
+        possible_filenames = (custom_config_file)
```
Fizzadar: Missing comma on the end here?

Fizzadar (approved):
One bug (I think), otherwise LGTM!

kubetools/config.py
stroebs: Good catch, fixed.

stroebs:
@Fizzadar I added the flag to the `config` command too for parity.

Fizzadar (approved):

PR #53, opened 2020-08-18
Improve the error message when current context is missing.

Fizzadar:
Unfortunately the issue is in the python k8s library which currently
raises an exception in this case rather than just setting no current
context value.

See: https://github.com/kubernetes-client/python/issues/1193


risktoparkrageram (approved):
cool!

Fizzadar:
Thanks @risktoparkrageram!

PR #55, opened 2020-09-03
Allow any k8s container config

### LINKED ISSUE #50: Support more/all K8s config options for deployment containers
Ideally in a way that does not require allowing certain fields, but simply uses everything.

Ie - in dev we take K8s config and -> compose (when using compose dev backend). In prod we basically pass the object as-is, with any additions (envars/etc).

Again, would remain all configured in a single `kubetools.yml` file (for now?).

Fizzadar:
Closes: #50

### DIFF
--- a/kubetools/cli/generate_config.py
+++ b/kubetools/cli/generate_config.py
@@ -3,12 +3,21 @@ import json
 from collections import defaultdict
 
 import click
+import yaml
 
 from kubetools.config import load_kubetools_config
 from kubetools.kubernetes.api import get_object_name
 from kubetools.kubernetes.config import generate_kubernetes_configs_for_project
 
 from . import cli_bootstrap
+
+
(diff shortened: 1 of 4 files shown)


gchazot (approved):
LGTM
I have a feeling this is not a backwards compatible change -> should be a major version bump.

As in, before this, most parts of `kubetools.yml` not understood by kubetools would be ignored while now they'll be passed to k8s/compose.

Fizzadar:
Agreed @gchazot - there's potential for invalid data to have been embedded inside container config that was previously ignored, so definitely a major bump for that 👍

ryan109 (approved):

stroebs (approved):

Fizzadar:
Thanks for the reviews @gchazot @ryan109 @stroebs!

PR #68, opened 2020-10-06
Fix nested kubetools dirs

### LINKED ISSUE #14: Running `ktd` in a subdirectory duplicates `.kubetools` directory
Since allowing `ktd` to run in a subdirectory of the repo, the `.kubetools` directory is now being created in the directory where we're running `ktd`, rather than the root directory. I think this should always stay as the root directory of the project, next to `kubetools.yml`.

Fizzadar:
Closes: #14

### DIFF
--- a/kubetools/config.py
+++ b/kubetools/config.py
@@ -99,6 +99,7 @@ def load_kubetools_config(
         ).format(' ({0})'.format(app_name) if app_name else ''))
 
     config = yaml.safe_load(config)
+    config['_filename'] = filename
 
     # Check Kubetools version?
     if 'minKubetoolsVersion' in config:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ### Unreleased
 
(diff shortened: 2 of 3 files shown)


kubetools/config.py
```diff
@@ -99,6 +99,7 @@ def load_kubetools_config(
         ).format(' ({0})'.format(app_name) if app_name else ''))
 
     config = yaml.safe_load(config)
+    config['_filename'] = filename
```
stroebs, line 102: Should this not be popped off for k8s, or are underscore keys ignored?

stroebs (approved):

kubetools/config.py
Fizzadar: No top level keys map to K8s objects luckily, that's all nested under `deployments`, `dependencies` & `upgrades`! Verified by the new tests (https://github.com/EDITD/kubetools/tree/master/tests/configs/basic_app) :)

Fizzadar:
Thanks for the review @stroebs!

PR #72, opened 2020-11-03
Correlate dev jobs with container context

gchazot:
Previous code used the build/dockerfile to find a container matching matching the
one wanted to run a job. However this is not accurate since:
* a dockerfile could be used by different containers
* a container may not have a dockerfile if it's running of a standard image

Using containerContext is a little more accurate since it describes more
precisely the container than just the dockerfile. Also any container with
a dockerfile will anyway require a containercontext.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,3 +16,14 @@ install: 'pip install -e .[dev]'
 script:
   - flake8
   - pytest --cov
+
+deploy:
+  provider: pypi
+  user: __token__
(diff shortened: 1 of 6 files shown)


Fizzadar (approved):
Much neater! LGTM

Fizzadar (approved):
👍 much improved!

Fizzadar (approved):

gchazot:
Thanks @Fizzadar

PR #83, opened 2021-05-07
Translate k8s command/args to compose entrypoint/command

gchazot:
## Purpose of PR
Fix bug where local dev with `ktd` wouldn't work if the docker image has an entrypoint and only "args" are passed through the kubetools config.

:warning: This looks like it would break backwards compatibility but in the direction of a simpler `kubetools.yml`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 ### Unreleased
 
+# v13.0.0
+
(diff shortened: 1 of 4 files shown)


Fizzadar (approved):

Fizzadar:
Missed the comment - definitely a good move here for a simpler configuration, no issues bumping major version for the breaking change IMO.

PR #88, opened 2021-06-24
Update kubernetes api create job with non-blocking option

### LINKED ISSUE #87: Add non-blocking job creation in kubetools
Currently create_job function in kubernetes api waits for job to have succeeded.

For longer running jobs (like crawls) it would be useful to have a creation option that does not block the caller for duration of job. I am not familiar with the status transitions here but ideally it would wait for job to be 'started' for some definition of started.

eatoncns:
## Purpose of PR

Currently the create job function waits for job to complete. As part of moving crawlers to jobs in kubernetes cluster we want to be able to kick off long running job and not wait for completion.

I looked at waiting for active field in job being > 1 but this seems to be more related to pod existing than being available and running so not sure it adds much value

Closes #87

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ### Unreleased
 
+# v13.1.0
+
+- Update kubernetes api create job with non-blocking option
+
 # v13.0.0
 
 - Translate k8s-style `command`/`args` options to docker[-compose]-style `entrypoint`/`command`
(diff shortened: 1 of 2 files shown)


Fizzadar (approved):

ryan109 (approved):
Does the create job function make sure the pod is available and the command was executed on it first - Is it possible to do that?

eatoncns:
> Does the create job function make sure the pod is available and the command was executed on it first - Is it possible to do that?

What we look at for job completion is the [status](https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/V1JobStatus.md) for pulled job. The only relevant thing I could see there was active number of running pods but when I tested it seemed to go to 1 before pod is available

Although looking at the CLI output there is a job status of "running"... will see what that corresponds to

eatoncns:
I checked the job status conditions and there is a single transition from `None` to type `Complete` when it succeeds :(

PR #89, opened 2021-06-29
Add function to kubernetes api to list running jobs

eatoncns:
This was in `crawl_controller` PR but extracting bits that felt more at home here


ryan109 (approved):
LGTM, need changelog to release!

PR #92, opened 2021-07-02
Check for environment variable override of settings file location

eatoncns:
## Purpose of PR

Crawl controller was updated to set environment variable pointing at kubetools config file. I am not sure where I saw `$KUBECONFIG` previously (maybe an older version!?) but we seem to have lost support for it


eatoncns:
Mistook which config file was the issue

PR #95, opened 2021-07-06
Fix ktd test / running of commands in containers

### LINKED ISSUE #85: command/args pattern breaks for migrations
v13.0.0 (https://github.com/EDITD/kubetools/pull/83) breaks upgrade commands in the `kubetools.yml`. 

Running as:
```
upgrades:
  - name: Migrate database
    containerContext: webserver
    command: [flask, db, upgrade]
```
causes kubetools to interpret the command as an entrypoint instead of performing the migration. However running as:
```
upgrades:
  - name: Migrate database
    containerContext: webserver
    args: [flask, db, upgrade]
```
fails wth:
```
File ".../lib/python3.9/site-packages/kubetools/dev/environment.py", line 184, in up
    ' '.join(upgrade['command']),
'command' not found
```

### LINKED ISSUE #94: Ensure repos work with latest kubetools version
- Dressing room
- Hydra

Looking at the code the issue is mainly around `ktd test`. As the change [here](https://github.com/EDITD/kubetools/commit/f842ab68ab39b4ad7d9682972e7b110d88d1eda7) converts `command` to docker-compose `entrypoint` there is no longer any way to run the test command in the container.

eatoncns:
## Purpose of PR

In v13 commands in kubetools config are translated to entrypoint in docker compose. When an entrypoint is present it is always run in container, with any command being passed to it as argument.

This means that currently running `ktd test` will run any backend start command present rather than the provided test command.

N.B. This is a breaking change. The only repo I see in github search effected is [here](https://github.com/EDITD/docs/blob/da424553457ebe5a263c36e58838ddd737349c4d/manifests/es_knn_exporter/kubetools.yml)

### DIFF
--- a/kubetools/dev/backends/docker_compose/config.py
+++ b/kubetools/dev/backends/docker_compose/config.py
@@ -123,13 +123,11 @@ def _create_compose_service(kubetools_config, name, config, envvars=None):
 
     service.update(config)
 
-    if 'args' in service:
-        if 'command' in service:
-            service['command'].extend(service.pop('args'))
-        else:
-            # Translate k8s 'args' to docker-compose 'command' that will be
-            # passed to entrypoint if it exists
-            service['command'] = service.pop('args')
+    # Translate k8s command/args to docker-compose entrypoint/command
+    if 'command' in service:
+        service['entrypoint'] = service.pop('command')
+    if 'args' in config:
+        service['command'] = service.pop('args')
 
     if 'build' in service and 'context' not in service['build']:
         service['build']['context'] = '.'
--- a/kubetools/dev/backends/docker_compose/config.py
+++ b/kubetools/dev/backends/docker_compose/config.py
@@ -123,11 +123,13 @@ def _create_compose_service(kubetools_config, name, config, envvars=None):
 
     service.update(config)
 
-    # Translate k8s command/args to docker-compose entrypoint/command
-    if 'command' in service:
-        service['entrypoint'] = service.pop('command')
-    if 'args' in config:
-        service['command'] = service.pop('args')
+    if 'args' in service:
+        if 'command' in service:
+            service['command'].extend(service.pop('args'))
+        else:
+            # Translate k8s 'args' to docker-compose 'command' that will be
+            # passed to entrypoint if it exists
+            service['command'] = service.pop('args')
 
     if 'build' in service and 'context' not in service['build']:
         service['build']['context'] = '.'
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ### Unreleased
 
+# v13.4.1
+
+- Fix regression from 13.0.0 causing upgrades and tests to fail when using an image with an entrypoint, or specifying
+  one with "command" in the config
+
 # v13.4.0
 
 - Expose propagation_policy param in k8s api job deletion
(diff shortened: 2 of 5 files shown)


kubetools/dev/backends/docker_compose/config.py
```diff
@@ -123,11 +123,13 @@ def _create_compose_service(kubetools_config, name, config, envvars=None):
 
     service.update(config)
 
-    # Translate k8s command/args to docker-compose entrypoint/command
-    if 'command' in service:
-        service['entrypoint'] = service.pop('command')
-    if 'args' in config:
-        service['command'] = service.pop('args')
+    if 'args' in service:
+        if 'command' in service:
```
eatoncns: There is an assumption here that args without command means entrypoint in docker file.

I did wonder whether args with command is a reasonable state to be in? I guess something like a command in main block and additional args in `dev` block might make sense

gchazot: yes, args + command is a valid situation.

In k8s specification, `command` is the ability to override Docker image's `entrypoint` and `args` is docker-compose's `command`

It makes things a little confusing hence this code that was making sure `kubetools.yml` is written with the k8s convention and translated to the docker-compose convention for development

gchazot (changes requested):
This is almost a straight revert of https://github.com/EDITD/kubetools/pull/83 (which is the breaking change between v12 and v13)

I don't think this is the right place to fix the problem here. I haven't had time to investigate much yet but I think the root cause lies where the `ktd test` command builds the docker/compose command to run.

gchazot:
> N.B. This is a breaking change. The only repo I see in github search effected is here

Indeed, v13 was created exactly for that kubetools file

eatoncns:
> This is almost a straight revert of #83 (which is the breaking change between v12 and v13)
> 
> I don't think this is the right place to fix the problem here. I haven't had time to investigate much yet but I think the root cause lies where the `ktd test` command builds the docker/compose command to run.

I think there is a general issue with setting an entrypoint in the docker compose config in that it means any command given just becomes params to the entrypoint i.e. running arbitrary command in the container like ktd test or migrations do not work

My thinking was to enable args to be given to an entrypoint in dockerfile with the `args` -> `command` translation (as in the updated example) but not lock out other situations

So [in the original problem situation](https://github.com/EDITD/docs/blob/da424553457ebe5a263c36e58838ddd737349c4d/manifests/es_knn_exporter/kubetools.yml) we could now just delete the `dev` block entirely and the args will be translated to command correctly for docker compose

kubetools/dev/backends/docker_compose/config.py
eatoncns: Yes, what I was wondering is whether there is a k8s case where it makes sense to specify both `command` and `args` instead of just the full command. The one I could think of is specifying a command in the normal kubeconfig block and then having some local specific argument in a `dev` block

gchazot: There's also a case of keeping separate the semantics of the main command of the container and the specific arguments to apply to the use-case.
Within kubetools context that could also mean a `command` in the containerContext (i.e. applying to all containers using it) and different `args` in each container that uses that context, including "one-shot" containers like the ones used for migrations or tests.

An example would be a django app with the containercontext describing how to build the container, base env variables and the main command (`manage.py`) then each container would run the same command with different args (runserver, migrate, test) ... Not a great example but explains the idea

gchazot (approved):

PR #97, opened 2021-07-08
Expose propagation_policy param in k8s api job deletion

### LINKED ISSUE #96: Look at propagation_policy in job deletion
Currently when calling kubernetes api `delete_job` the job object is deleted buts pods are left running.

From doc string in corresponding kubernetes api there is an option for `propagation_policy` which we might want to expose in order to control this behaviour.

eatoncns:
## Purpose of PR

This parameter allows the caller to specify whether to orphan or kill
dependents. The latter can be done in background (async) or foreground.

Introducing as default config in one of our use cases is orphan and we
would rather have spawned pods removed to actually stop job processing.

### DIFF
--- a/kubetools/kubernetes/api.py
+++ b/kubetools/kubernetes/api.py
@@ -274,11 +274,17 @@ def list_running_jobs(env, namespace):
     return [job for job in jobs if is_running(job)]
 
 
-def delete_job(env, namespace, job):
+valid_propagation_policies = ["Orphan", "Background", "Foreground"]
+
+
+def delete_job(env, namespace, job, propagation_policy=None):
+    if propagation_policy and propagation_policy not in valid_propagation_policies:
+        raise KubeBuildError(f"Propagation policy must be one of {valid_propagation_policies}")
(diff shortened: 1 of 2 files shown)


kubetools/kubernetes/api.py
```diff
@@ -274,11 +274,17 @@ def list_running_jobs(env, namespace):
     return [job for job in jobs if is_running(job)]
 
 
-def delete_job(env, namespace, job):
+valid_propagation_policies = ["Orphan", "Background", "Foreground"]
+
+
+def delete_job(env, namespace, job, propagation_policy=None):
+    if propagation_policy and propagation_policy not in valid_propagation_policies:
```
eatoncns, line 281: I felt like some sign posting of potential values was useful here but the default is a bit tricky as it is config dependent. The underlying api does ignore `None` values though so this does maintain defaults correctly. A _bit_ reliant on client implementation but I thought the added 'documentation' was worth the associated risk

Fizzadar (approved):

PR #99, opened 2021-07-14
Update push command to allow additional tags for image

### LINKED ISSUE #98: Update push command to allow additional tags

eatoncns:
## Purpose of PR

This allows CLI user to pass tags for image that will be pushed as well as the default commit hash based tag.

image

### DIFF
--- a/kubetools/deploy/image.py
+++ b/kubetools/deploy/image.py
@@ -14,15 +14,19 @@ def get_commit_hash_tag(context_name, commit_hash):
     return '-'.join((context_name, 'commit', commit_hash))
 
 
-def get_docker_tag(registry, app_name, context_name, commit_hash):
-    # Tag the image like registry/app:commit-hash
-    docker_version = '{0}:{1}'.format(
-        app_name,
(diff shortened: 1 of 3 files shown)


Fizzadar (approved):

kubetools/deploy/image.py
```diff
@@ -84,7 +88,10 @@ def _ensure_docker_images(
     kubetools_config, build, app_dir, commit_hash,
     default_registry=None,
     check_build_control=lambda build: None,
+    additional_tags=None,
 ):
+    if additional_tags is None:
+        additional_tags = []
```
ryan109, line 94: Could we just take in `additional_tags` as an empty list as kwarg default? Or is that unpythonic of me

ryan109 (approved):

kubetools/deploy/image.py
eatoncns: In this case that would be fine as it is read only. I went this way out of habit, as mutable defaults are a pain when they are mutated in the function.

PR #103, opened 2021-07-19
Update to only build/push docker images relevant to current context

### LINKED ISSUE #104: Ensure kubetools push only builds docker images relevant to current context

eatoncns:
## Purpose of PR

Currently when ensuring images are present in repo during a deploy or push we loop over all container contexts present in the config. As context filters (dev, envs, namespaces etc.) are not applied to container contexts this results in unnecessary images being built/pushed.

This change updates to look only at containers that are present in config after filtering has been applied. Container contexts are then looked up from there.

### DIFF
--- a/kubetools/deploy/image.py
+++ b/kubetools/deploy/image.py
@@ -49,19 +49,20 @@ def has_app_commit_image(registry, app_name, context_name, commit_hash):
 
 
 def _get_container_contexts_from_config(app_config):
-    context_name_to_build = {
-        key: context['build']
-        for key, context in app_config.get('containerContexts', {}).items()
-        if 'build' in context
-    }
-
+    context_name_to_build = {}
+    container_contexts = app_config.get('containerContexts', {})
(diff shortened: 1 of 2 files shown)


kubetools/deploy/image.py
```diff
@@ -49,19 +49,16 @@ def has_app_commit_image(registry, app_name, context_name, commit_hash):
...
-
+    context_name_to_build = {}
+    container_contexts = app_config.get('containerContexts', {})
     for deployment, data in app_config.get('deployments', {}).items():
         containers = data.get('containers')
         for name, container in containers.items():
             if 'containerContext' in container:
-                continue
+                context_name = container['containerContext']
+                context_name_to_build[context_name] = container_contexts[context_name]['build']
```
eatoncns: Realising this doesn't account for the no build key case that was handled in the previous code. Will update

ryan109 (approved):

PR #111, opened 2021-07-28
Add tests for building run compose command

eatoncns:
## Purpose of PR

As there have been a few changes for various cases these should help serve as documentation

### DIFF
--- a/tests/test_build_compose_command.py
+++ b/tests/test_build_compose_command.py
@@ -1,0 +1,45 @@
+from unittest import TestCase
+
+from kubetools.dev.backends.docker_compose import build_run_compose_command
+
+
+class TestBuildComposeCommand(TestCase):
+    def setUp(self):
+        self.container = 'container'
+
+    def test_simple_composite_command(self):
+        command = ('python', 'scripts/blah.py', '--argument')
+        result = build_run_compose_command(self.container, command, envvars=None)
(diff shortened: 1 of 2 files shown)


tests/test_build_compose_command.py
```diff
@@ -0,0 +1,45 @@
...
+        expected_result = ['run', '--entrypoint', 'python scripts/blah.py --argument', 'container']
+        self.assertEqual(expected_result, result)
+
+    def test_composite_command_is_shell_escaped(self):
+        command = ('python', 'scripts/blah.py', '--argument', 'key: value')
+        result = build_run_compose_command(self.container, command, envvars=None)
+        expected_result = [
+            'run',
+            '--entrypoint',
+            "python scripts/blah.py --argument 'key: value'",
```
ryan109, line 28: is it shlex that knows to keep `'key: value'` in quotations? How's it done that, they are all just separate strings on input?

ryan109 (approved):

tests/test_build_compose_command.py
eatoncns: Yep, if there are any spaces in the thing shlex is quoting it wraps the whole thing in quotes. This is what broke the simple single command case for megabucks.

eatoncns: So in this case the input to shlex is just the `'key: value'`. That gets output as `"'key: value'"` which means the inner quotes are preserved when the whole command is joined together

PR #118, opened 2021-12-31
Add cron job object

DilaraOflaz:
**## Purpose of PR**
I added CronJob object to Kubetools.

**## Parts of the app this will impact**
With this Branch we can run the following commands:
- `kubetools deploy ...`
- `kubetools remove ...`
- `kubetools show ...`
- `kubetools cleanup ...`

**## How to test it**
1.  if you have a Mac, you can test it locally by enabling Kubernetes in Docker-Desktop ( Open Docker -> Preferences -> Kubernetes -> Check **Enable Kubernetes** -> Apply & Restart) Otherwise, by pointing to the **staging**  Kubernetes  cluster

2. To get the name of the context, run this following command: `kubectl config current-context`

3. Checkout to this branch, create a python Virtual Environment, run the following command: `pip install -e .[dev]`
_Note: Run `pip install -e ".[dev]"` if you use zsh_

4. Create a **kubetools.yml** file. Content example:

```
name: my-app
namespace: cronjob-test
cronjobs:
  my-cronjob:
    schedule: "*/1 * * * *"
    concurrency_policy: "Replace"
    containers:
      hello:
        image: busybox
        command: [/bin/sh, -c, date; echo Hello from the Kubernetes cluster]
```

5. Run this command: `kubetools --context <CONTEXT NAME> deploy cronjob-test`
This command can take 20s to finish running (it waits until the first job is created)

6. The `deploy` command should create the `cronjob-test` namespace and the `my-cronjob` CronJob, you can check by running this command: `kubetools --context <CONTEXT NAME> show cronjob-test`

7. Run this command to delete the CronJob: `kubetools --context <CONTEXT NAME> remove cronjob-test`
After running this command, if you run this following command `kubetools --context <CONTEXT NAME> show cronjob-test`, the output should be `Nothing to be found here 👀!`

8. To clean the succeeded/completed pods, you can run this following command:
`kubetools --context <CONTEXT NAME> cleanup cronjob-test`

**## Additional information**
With the current code, the `kubetools cleanup` command cleans only if the pods (created by the CronJob) are orphaned. If the CronJob still exists (Not removed), even if the pods are completed they are not cleaned.


.gitignore
```diff
@@ -6,3 +6,7 @@ build/
 
 docs/_html
 docs/build
+
+# TO REMOVE
+kubetools_env/
+kubetools.yml
```
eatoncns: 👀

tests/test_config_generation.py
```diff
@@ -24,7 +24,7 @@ def _test_configs(folder_name, **kwargs):
     kubetools_config = load_kubetools_config(app_dir, **kwargs)
 
     with mock.patch('kubetools.kubernetes.config.job.uuid4', lambda: 'UUID'):
-        services, deployments, jobs = generate_kubernetes_configs_for_project(
+        services, deployments, jobs, cronjobs = generate_kubernetes_configs_for_project(
```
eatoncns, line 27: Maybe due to the gitignore but I was expecting to see an update to one of the test kubetool configs to add cron?

tests/configs/basic_app/k8s_cronjobs.yml
```diff
@@ -0,0 +1,23 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  annotations: {app.kubernetes.io/managed-by: kubetools}
+  labels: {kubetools/name: generic-cronjob, kubetools/project_name: generic-app,
+    kubetools/role: cronjob}
+  name: generic-cronjob
+spec:
+  schedule: "*/1 * * * *"
+  startingDeadlineSeconds: 10
```
ryan109, line 16: This can be removed I think as we never interpret this, we set it to 10 by default in the cronjob creation.

ryan109: Unless we want to allow the user to configure `startingDeadlineSeconds`

kubetools/kubernetes/config/cronjob.py
```diff
@@ -0,0 +1,77 @@
...
+                            'labels': labels,
+                            'annotations': annotations,
+                        },
+                        'spec': {
+                            'restartPolicy': 'OnFailure',
+                            'containers': kubernetes_containers,
+                        },
+                    },
+                },
+            },
```
ryan109: We should to add a line for `concurrencyPolicy` [configuration], it would be useful to be able to stop new cronjobs from cancelling out old ones that are long running (https://kubernetes.io/docs/tasks/job/automated-tasks-with-cron-jobs/#concurrency-policy)

```suggestion
        'spec': {
            'schedule': schedule,
            'startingDeadlineSeconds': 10,
            'concurrencyPolicy': concurrency_policy,
            'jobTemplate': {
                'spec': {
                    'template': {
                        'metadata': {
                            'name': cronjob_name,
                            'labels': labels,
                            'annotations': annotations,
                        },
                        'spec': {
                            'restartPolicy': 'OnFailure',
                            'containers': kubernetes_containers,
                        },
                    },
                },
            },
```

kubetools/deploy/commands/deploy.py
```diff
@@ -245,3 +271,13 @@ def execute_deploy(build, namespace, services, deployments, jobs, delete_complet
             for service in exist_main_services:
                 build.log_info(f'Update service: {get_object_name(service)}')
                 update_service(build.env, build.namespace, service)
+
+    for cronjob in cronjobs:
+        with build.stage('Create and/or update cronjobs'):
+            if cronjob['metadata']['labels'][ROLE_LABEL_KEY] == 'cronjob':
```
ryan109: Trying to understand the need for this label check, is it possible for a cronjob object to be passed into `execute_deploy()` that doesn't have the label `kubetools/role: cronjob`? If so, it's worth having an else to warn the user why:
```
else:
  build.log_info(f'Not creating/updating {get_object_name(cronjob)} as label: "kubetools/role" is incorrect')
```
If that's the use case we want from this

kubetools/deploy/commands/remove.py
```diff
@@ -58,3 +68,7 @@ def execute_remove(build, services, deployments, jobs):
     if jobs:
         with build.stage('Delete jobs'):
             delete_objects(build, jobs, delete_job)
+
+    if cronjobs:
+        with build.stage('Delete cronjobs'):
+            delete_objects(build, cronjobs, delete_cronjob)
```
ryan109, line 76: From what I can see this will delete all cronjobs associated with a project based on the filtering done [here](https://github.com/EDITD/kubetools/blob/f69e211e664d99376437d8db682c1e92581f9744/kubetools/deploy/util.py#L74-L88) - think it involves a fair bit more work to produce a UI that can only delete individual jobs, but worth looking into in the future. OOS for this PR.

ryan109 (commented):
Looking great, a couple of things for discussion/changes :)

tests/configs/basic_app/k8s_cronjobs.yml
ryan109: But this is a test case, but could be confusing as if I looked at the test case I would think we could edit `startingDeadlineSeconds`

.gitignore
DilaraOflaz: Removed

tests/configs/basic_app/k8s_cronjobs.yml
DilaraOflaz: I removed `startingDeadlineSeconds`

kubetools/kubernetes/config/cronjob.py
DilaraOflaz: I added the `concurrencyPolicy` configuration

kubetools/deploy/commands/deploy.py
DilaraOflaz: This is a bad copy/paste, I removed it

kubetools/deploy/commands/remove.py
DilaraOflaz: I added a comment to not forget

tests/test_config_generation.py
DilaraOflaz: I added the following test file `tests/configs/basic_app/k8s_cronjobs.yml`

ryan109:
Before we merge this can we squash a lot of these commits? For example a lot of the "cleaning", and "testing" ones could be pushed into one or merged with feature commits. Sam will be covering exactly this in the git meeting tomorrow, but here's a general tutorial on it: https://www.baeldung.com/ops/git-squash-commits

It keeps the commit tree in a readable state when looking through history, so a commit should be named exactly as it is "Adding functionality X to support feature Y" - as an example.

Let me know if you have questions about it or want to go over it together.

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ### Unreleased
 
+# v13.8.2
+- Add CronJob object
```
gchazot: I think it's worth bumping the minor version here, rather than just the patch version. This is a major feature addition

tests/configs/basic_app/k8s_cronjobs.yml
```diff
@@ -0,0 +1,23 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  annotations: {app.kubernetes.io/managed-by: kubetools}
+  labels: {kubetools/name: generic-cronjob, kubetools/project_name: generic-app,
+    kubetools/role: cronjob}
```
gchazot:
```suggestion
  labels: {
    kubetools/name: generic-cronjob,
    kubetools/project_name: generic-app,
    kubetools/role: cronjob
  }
```
This format would be more readable or does it break the test?

tests/configs/basic_app/k8s_cronjobs.yml
```diff
@@ -0,0 +1,23 @@
...
+  name: generic-cronjob
+spec:
+  schedule: "*/1 * * * *"
+  concurrency_policy: "Replace"
+  suspend: false
+  jobTemplate:
+    spec:
+      template:
+        labels: {kubetools/name: generic-cronjob, kubetools/project_name: generic-app,
+          kubetools/role: cronjob}
```
gchazot:
```suggestion
        labels: {
          kubetools/name: generic-cronjob,
          kubetools/project_name: generic-app,
          kubetools/role: cronjob
        }
```
Same format suggestion if it works

tests/test_config_generation.py
```diff
@@ -34,6 +34,8 @@ def _test_configs(folder_name, **kwargs):
         _assert_yaml_objects(deployments, path.join(app_dir, 'k8s_deployments.yml'))
     if jobs:
         _assert_yaml_objects(jobs, path.join(app_dir, 'k8s_jobs.yml'))
+    if cronjobs:
+        _assert_yaml_objects(cronjobs, path.join(app_dir, 'k8s_cronjobs.yml'))
```
gchazot: Sorry I don't understand how this test file is validating the change:
* I think there should be also a change in `tests/configs/basic_app/kubetools.yml` to add a cronjob
* This test function could also be improved to look for all `k8s_blah.yml` files in the test folder and do the assertion if present. I mean. here we expect a k8s_cronjob.yml to be generated, but we only assert its presence if the kubetools config contains cronjobs. The solution could be to change `if cronjobs:` into `if cronjobs or os.path.isfile(path.join(app_dir, 'k8s_cronjobs.yml'))` + a little refactor to make it all more readable

DilaraOflaz:
I found this parameter `'ttlSecondsAfterFinished': X`, to remove the pods automatically `X` seconds after completion

I can add it inside the file `kubetools/kubernetes/config/cronjob.py` inside the cronjob spec: `spec.jobTemplate.spec` 

So Kubernetes will handle the deletion of the completed pods, we won't have to modify the `delete_cronjob` function

Please let me know your thoughts ?

edmond-edited:
PR looks good to me, I think we maybe we should also consider these parameters for job history retention `spec.successfulJobsHistoryLimit` and `spec.failedJobsHistoryLimit`? By default, `successfulJobsHistoryLimit` is set to 3 and `failedJobsHistoryLimit` is set to 1 for Kubernetes

kubetools/cli/show.py
```diff
@@ -98,6 +99,17 @@ def _get_command(item):
...
+    # CronJob didn't start yet
+    if item.status.last_schedule_time is None and item.status.last_successful_time is None:
+        return '0/1'
+    # CronJob is running
+    elif item.status.last_schedule_time is not None:
+        return '1/1'
+    else:
+        return 'unknown'
+
+
```
gchazot: Clean code: Organise code for reading top-to-bottom, so that the position in the file reflects function nesting.

This means here this function should be located further down the file, **after** the place where it's used, which here means the end of the file

kubetools/cli/show.py
```diff
@@ -98,6 +99,17 @@ def _get_command(item):
...
 
+def _get_cronjob_status(item):
+    # CronJob didn't start yet
+    if item.status.last_schedule_time is None and item.status.last_successful_time is None:
+        return '0/1'
+    # CronJob is running
+    elif item.status.last_schedule_time is not None:
+        return '1/1'
+    else:
+        return 'unknown'
```
gchazot: This does not seem to be consistent with the CronJob API documentation
https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/cron-job-v1/#CronJobStatus

* I think "CronJob didn't start yet" should only be based on `item.status.last_schedule_time`, because I would expect that if `item.status.last_schedule_time is None` then `item.status.last_successful_time` will always be `None`.
* There is another value `item.status.active` that contains a list of **active** Jobs scheduled by the CronJob. I suspect this is what we should be using for "CronJob is running"

kubetools/deploy/commands/deploy.py
```diff
@@ -124,7 +130,14 @@ def log_deploy_changes(
...
         for deployment in list_deployments(build.env, build.namespace)
     )
+    existing_cronjobs_names = set(
+        get_object_name(cronjob)
+        for cronjob in list_cronjobs(build.env, build.namespace)
+    )
 
+    deploy_cronjobs_names = set(
+        get_object_name(cronjob) for cronjob in cronjobs
+    )
```
gchazot: For consistency, this should be moved after `deploy_deployment_names = ...` below

kubetools/kubernetes/config/__init__.py
```diff
@@ -269,4 +270,41 @@ def generate_kubernetes_configs_for_project(
...
+        schedule = cronjob['schedule']
+        concurrency_policy = cronjob['concurrency_policy']
+
+        if container_ports:
+            services.append(make_service_config(
+                name,
+                container_ports,
+                labels=cronjob_labels,
+                annotations=app_annotations,
+            ))
```
gchazot: Would it make sense to have ports published as services by a cronjob? If yes, then why not for a regular job?

gchazot (commented):
Looks good to me overall.
A few sugestions and questions.

Also I don't think that the test validates anything in the current state

CHANGELOG.md
DilaraOflaz: Yes, done

kubetools/deploy/commands/deploy.py
DilaraOflaz: Done

kubetools/cli/show.py
DilaraOflaz: done

tests/configs/basic_app/k8s_cronjobs.yml
DilaraOflaz: Done

tests/configs/basic_app/k8s_cronjobs.yml
DilaraOflaz: Done

kubetools/kubernetes/config/__init__.py
DilaraOflaz: Not really I removed it

kubetools/cli/show.py
DilaraOflaz: I've changed the `_get_cronjob_status` function

Here is the commit link: https://github.com/EDITD/kubetools/pull/118/commits/04c884fb84cfbd4662dd6fc2b550ea3e4ca653b4

Please let me know what you think ?

kubetools/kubernetes/config/__init__.py
gchazot: c12a2ee

CHANGELOG.md
gchazot: 06faa14c0b4f71226426110b30571a8ebf293760

kubetools/cli/show.py
gchazot: 7f3c135a16930c0508371fab4c19197f91ebd507

kubetools/deploy/commands/deploy.py
gchazot: 7f3c135a16930c0508371fab4c19197f91ebd507

tests/configs/basic_app/k8s_cronjobs.yml
gchazot: 7f3c135a16930c0508371fab4c19197f91ebd507

tests/configs/basic_app/k8s_cronjobs.yml
gchazot: 7f3c135a16930c0508371fab4c19197f91ebd507

tests/test_config_generation.py
DilaraOflaz: I added a cronjob inside the following file `tests/configs/basic_app/kubetools.yml`. I also corrected the file `tests/configs/basic_app/k8s_cronjobs.yml`.

Here is the commit link: https://github.com/EDITD/kubetools/pull/118/commits/df9cad1fd598ed3298f0851350bb9b08507f5570

kubetools/cli/show.py
gchazot: I'm not clear yet how this works, or even what this code is used for. Have you tested it? What command did you use to test?

DilaraOflaz: Here is the code that I added:
```
def _get_cronjob_status(item):
    # CronJob is running
    if item.status.active is not None:
        return '1/1'
    else:
        # CronJob didn't start yet
        if item.status.last_schedule_time is None:
            return '0/1'
        # CronJob is running but the Job didn't start yet
        else:
            return '1/1'
```

From what I've seen, the `active` parameter in not `None` if there is a job (of this Cronjob) currently running.
So for example if you run the command: `kubetools deploy cronjob-tests`
and then if you run the command `kubetools show cronjob-test`

On the `Ready` column of the Cronjob line you will see `1/1` or `0/1` (this is the part that this function is changing, you can change the function to see how it works)

One way to see how the `Ready` column changes, it's to add a print between line `209` and `210`:
```
def _get_cronjob_status(item):
    print(f"{item.status}")
    # CronJob is running
```

So to explain the code of the function `_get_cronjob_status`:
```
if  item.status.active is not None --> Job of Cronjob is currently running
else:
    if item.status.last_schedule_time is None --> 0 Job has been ran yet, Cronjob is not ready
    else --> there is a job of Cronjob that has ran and finished, so waiting for next job to start (Cronjob is running but doesn't appear active, because there isn't a Job of this Cronjob running at this moment)
```

tests/configs/basic_app/kubetools.yml
```diff
@@ -25,3 +25,12 @@ deployments:
...
             path: /ping
+
+cronjobs:
+  generic-cronjob:
+    schedule: "*/1 * * * *"
+    concurrency_policy: "Allow"
+    containers:
+      generic-container:
+        image: generic-image
+        command: [generic-command]
```
gchazot: This would be more in line with the spirit of this test case
```suggestion
    containers:
      generic-container:
        containerContext: generic-context
```

tests/test_config_generation.py
gchazot: * df9cad1 covers the first point. Adding a comment on the test file
* the second point is not covered yet

tests/configs/basic_app/kubetools.yml
DilaraOflaz: Done, here is the commit link https://github.com/EDITD/kubetools/pull/118/commits/abf557fb47b73c192123c2667795be92cbe6e4d7

tests/test_config_generation.py
DilaraOflaz: For the second point, I don't really understand because we are comparing the objects (services, deployments, jobs, cronjobs) with their respective config files

So if the object doesn't exist, this won't work.

We have to check with `and` and not `or`: `if cronjobs and os.path.isfile(path.join(app_dir, 'k8s_cronjobs.yml'))`

To be sure that both exist, do you agree ?

DilaraOflaz: Done, here is the commit link: https://github.com/EDITD/kubetools/pull/118/commits/035611dd5c8def476ca9f45f72dba68a9f41755e

tests/configs/basic_app/kubetools.yml
gchazot: image
Typical example where a functional code change has been added in an unrelated commit

gchazot: Also, why did you create a new `containerContext` rather than re-using the existing one?
The only difference is that the ports are not in the new containerContext, but IIRC you coded it so CronJobs ignore ports anyway...

tests/test_config_generation.py
gchazot: Awesome! Thank you!

tests/configs/basic_app/kubetools.yml
DilaraOflaz: For this change:
```
    'upgrades',
     'deployments',
     'dependencies',
     'cronjobs',
```

I added the `cronjobs` object because otherwise it doesn't link the `containerContext` inside the `cronjobs` object to the `containerContext` that is inside the list of `containerContext`

So it's link to the commit in the sens of, if I don't add this small change the `pytest --cov` wouldn't work

DilaraOflaz: I added the second `containerContext` because of the port, but I removed it

Here is the commit: https://github.com/EDITD/kubetools/pull/118/commits/07b3c91a4c963902c0164f3f02e7b6e615a99d88

kubetools/cli/show.py
gchazot: OK, makes more sense now. Then I'd change the output to be:
```
if  item.status.active is not None:
    Job of Cronjob is currently running
    -> "0/1"
elif item.status.last_schedule_time is None:
    0 Job has been ran yet, Cronjob is not ready
    -> "0/0"
else:
    there is a job of Cronjob that has ran and finished, so waiting for next job to start (Cronjob is running but doesn't appear active, because there isn't a Job of this Cronjob running at this moment)
    -> "1/1"

Comment (DilaraOflaz) [reply kubetools/cli/show.py]: One question

Why do we put  `-> "0/1"` if active  `is not None` ?
If active is not None that means that the Cronjob is up and there is a job running. It should be `-> "1/1"`

And in the `else:` part, it should be `-> "0/1"`
Because that means that the Cronjob is up but there is no job running at this moment.

I think it should be like below no ?

```
if  item.status.active is not None:
    Job of Cronjob is currently running
    -> "1/1"
elif item.status.last_schedule_time is None:
    0 Job has been ran yet, Cronjob is not ready
    -> "0/0"
else:
    there is a job of Cronjob that has ran and finished, so waiting for next job to start (Cronjob is running but doesn't appear active, because there isn't a Job of this Cronjob running at this moment)
    -> "0/1"
```

Comment (gchazot) [reply kubetools/cli/show.py]: My idea is that the first digit is for the success of a job, while the second one is for the success of the job scheduling. Actually, I think I missed something and it should be:
```
if  item.status.active is not None:
    Job is currently running (implies successfully started)
    -> "?/1"
elif item.status.last_successful_time is not None:
   Job has been successful (implies successfully started)
    -> "1/1"
elif item.status.last_schedule_time is not None:
    Job has been scheduled
    -> "0/1"
else:
    Job has never been scheduled (error in CronJob?)
    -> "0/0"
```

Maybe using a word describing each situation would help.

Also, the [k8s documentation](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/) is not giving anything about the dependencies between the statuses of a CronJob. So I'm making the assumptions that:
* a Job cannot be running if it's not been scheduled
* a Job cannot be successful if it's not been scheduled
* a Job cannot be scheduled if the CronJob was not created in k8s


Let me know what you think, I might be missing something

tests/configs/basic_app/kubetools.yml
gchazot: >  I added the cronjobs object because otherwise it doesn't link the containerContext inside the cronjobs object to the containerContext that is inside the list of containerContext

Ok. Makes sense. The structure of the code is quite misleading, because it seems that the logic for "conditions" is driving the use of containerContext... this is quite confusing

gchazot: This conversation is :+1: for me. Will keep it open for this explanation^^

kubetools/cli/show.py
DilaraOflaz: Okay make sense! This looks good, I did the change

Here is the commit: https://github.com/EDITD/kubetools/pull/118/commits/1450de2fea4ebaa062bb98292aabd67e8ff500ad

gchazot (approved):
Great work @DilaraOflaz ! Thanks

ryan109 (approved):

ryan109:
Great stuff!

PR #121, opened 2022-03-28
pin docker-compose <1.29 as it requires docker v5

ryan109:
## Purpose of PR
At the moment if you install kubetools it will install `1.29.*`, which breaks due to:
```
pkg_resources.ContextualVersionConflict: (docker 4.4.4 (/Users/ryanmccorry/.pyenv/versions/3.7.6/lib/python3.7/site-packages), Requirement.parse('docker[ssh]>=5'), {'docker-compose'})
```
## Related links

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -59,7 +59,8 @@ if __name__ == '__main__':
             'kubernetes',
             'tabulate<1',
             # compose v2 has broken container naming
-            'docker-compose<2',
+            # and compose 1.29 requires docker >5
+            'docker-compose<1.29',
         ),
         extras_require={
             'dev': (


edmond-edited (commented):
Ignore, misread

edmond-edited (approved):

setup.py
```diff
@@ -59,7 +59,8 @@ def get_readme_content():
             'kubernetes',
             'tabulate<1',
             # compose v2 has broken container naming
-            'docker-compose<2',
+            # and compose 1.29 requires docker >5
```
mattalexanderhill, line 62: Is there any reason to specify `docker<5`? 

The only breaking changes listed are dropping python 2.7 and 3.5 support, are they deal breakers?

https://docker-py.readthedocs.io/en/stable/change-log.html

mattalexanderhill: This is where it was pinned last time: https://github.com/EDITD/kubetools/commit/3a668b40cd68c0b60b5f5d1922bd290c3cc14618

I think before `docker==5.0` was released https://pypi.org/project/docker/#history

gchazot (approved):

setup.py
ttecles: I guess this is a safety guard (pinning to a lower mayor version) but given that this is a package I would also try to loosen the dependencies as much as possible. And seems @mattalexanderhill is right.

@ryan109 @gchazot @edmond-edited what do you think about this?

gchazot: needs testing...
but otherwise no reason ~to pin docker~ **NOT** to pin docker

gchazot:
Is this still necessary? AFAIK kubetools is not causing any issues at the moment without it...

ryan109:
Honestly can't remember, I can close it for now and reopen if we see issues.

PR #126, opened 2022-06-13
enable jobs to pass in resources for containers

fmcwalters-edited:
in order to prevent jobs from hogging resources, we need to be able
to add limits. These are optional and can be passed in as a dictionary
adhering to:
https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/

## Related links


seangarner:
@fmcwalters-edited what's the philosophy here - should we be proactively setting limits in all projects, or only ones where we run into issues?

fmcwalters-edited:
@seangarner this is a reactive change required for Skuscraper as we had a few crawlers that would end up hogging RAM. I made `resources` a non-required variable as I didn't want to have to retroactively enforce this for every single project.  
Going forward it would be nice to have limits set on containers in projects, this enables us to do that; however it's beyond the scope of this PR to enforce this and I don't feel comfortable unilaterally forcing architectural changes on every single project that uses Kubetools and containers.

seangarner:
Thanks for the clarification @fmcwalters-edited.

It's made me think about the perspective of smart-proxy, given it's a node project and handles memory differently to python.  If you're curious, I've asked the question specific to smart-proxy [over in slack](https://edtd.slack.com/archives/C02H1M3JQ0P/p1655201329188889).

ryan109:
@seangarner in an ideal world every single project would have limits on it. Resource limits have a number of applications, but our two biggest benefits are:
- A hard cap, doesn't allow the container to go over the set amount of RAM or CPU usage (although from a chat with @gchazot this morning, apparently the CPU limits are a bit buggy in k8s). So if you set a limit of 2g and the container immediately comes up and try to init a package that requires 3, the container falls over and gets stuck in a crashloop.
- k8s can be more effective in balancing deployments across the cluster. If we have 10 deployments with proper CPU & RAM allocation, then it knows exactly what to expect and how its best to spread these across nodes. If we add an 11th deployment with no specification then k8s is essentially operating blind, as it could be super lightweight and not impact that much, or each pod might need 30g of RAM. It has no way of making those resource claims until the deployment comes up and it notices how much extra it needs, which we have to retroactively expand the cluster. If we had the limits in and the cluster was full, the deployment _should_ just fail.

Point 1) is useful regardless of how many projects we apply limits to, point 2) is only useful if _every_ deployment in the cluster has limits. Only our pipeline cluster effectively employs 2), but with these new scraping clusters we could set the precedence and get these benefits now. Hydra applies its limits, skuscraper is about to have limits applied, so if we done it for smart-proxy we would be fulfilling this k8s architecture dream.

**caveat**: VM shooter is also in the ingestion cluster, I don't think it has limits.

ryan109:
@seangarner in an ideal world every single project would have limits on it. Resource limits have a number of applications, but our two biggest benefits are:
- A hard cap, doesn't allow the container to go over the set amount of RAM or CPU usage (although from a chat with @gchazot this morning, apparently the CPU limits are a bit buggy in k8s). So if you set a limit of 2g and the container immediately comes up and try to init a package that requires 3, the container falls over and gets stuck in a crashloop.
- k8s can be more effective in balancing deployments across the cluster. If we have 10 deployments with proper CPU & RAM allocation, then it knows exactly what to expect and how its best to spread these across nodes. If we add an 11th deployment with no specification then k8s is essentially operating blind, as it could be super lightweight and not impact that much, or each pod might need 30g of RAM. It has no way of making those resource claims until the deployment comes up and it notices how much extra it needs, which we have to retroactively expand the cluster. If we had the limits in and the cluster was full, the deployment _should_ just fail.

Point 1) is useful regardless of how many projects we apply limits to, point 2) is only useful if _every_ deployment in the cluster has limits.

kubetools/kubernetes/config/container.py
```diff
@@ -117,6 +120,8 @@ def make_container_config(
                 'mountPath': volume.split(':')[1],
                 'name': get_hash(volume),
             })
+    if resources:
+        container_data['resources'] = resources
```
gchazot: This extra function argument is not necessary, it can be passed inside the `container` object ant will be copied over just below at line 122/127

gchazot (changes requested):
This can be simplified.

I'm wondering why we're not passing the whole `config` from the job to `make_container_config` but that change is out of scope.

kubetools/kubernetes/config/container.py
fmcwalters-edited: addressed at 316dc7f9596403086c6221b11b8f0d75f104e196

gchazot (changes requested):
I don't understand the purpose of this new parameter. None of the places that call this function pass any `resources` argument.

ryan109:
@gchazot the use case isn't within another kubetools function, but when we import the package in other services such as [here](https://github.com/EDITD/crawl_controller/blob/cf0cc319e41807887954666d83346e8c50aef351/crawl_controller/util/kubernetes.py#L54)

ryan109:
The idea being we can import kubetools inside another python service, this function will generate a kuberetes config using an input kubetools file and then call k8s directly to create the job with desired resources attached.

ryan109:
@fmcwalters-edited tests are failing as we're passing `'resources': None}` in (as no resources are specified in the test config file). We need to confirm k8s will accept `'resources': None`, if so then we need to update the test file/test to handle it.

gchazot:
From the point of view of `kubetools`, this is a half-baked functionality. Adding this as a possibility in the config generator means it also needs to be available from `kubetools.yml`.
`kubetools` is first a CLI tool.

ryan109:
@gchazot see [932b171](https://github.com/EDITD/kubetools/pull/126/commits/932b17194c08405d5e57c76de202c969418cc619) and [ce94a27](https://github.com/EDITD/kubetools/pull/126/commits/ce94a277b46ef162be59150f924bdeaab5647ad0).

Right now we can only create `job` objects from the "upgrade" functionality in `kubetools.yml`. There's a bit of work to be done if we wanted to add the ability to parse job objects the same way we do cronjobs/deployments. But that's OOS for this PR.

ryan109 (approved):

tests/configs/basic_app/k8s_jobs.yml
```diff
@@ -24,4 +24,7 @@ spec:
         image: generic-image
         imagePullPolicy: Always
         name: upgrade
+        resources: {
+          requests: { 'cpu': 1, 'memory': '1Gi'}
+        }
```
gchazot:
```suggestion
        resources:
          requests:
            cpu: 1
            memory: "1Gi"
```
I believe this is the correct formatting

gchazot (changes requested):

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ### Unreleased
 
+# v13.9.1
+- Allow jobs to specify Resources within Job spec
 # v13.9.0
```
gchazot:
```suggestion
# v13.9.1
- Allow jobs to specify Resources within Job spec

# v13.9.0
```

ryan109 (approved):

gchazot (approved):
LGTM

PR #127, opened 2022-06-16
Try to avoid duplicated dev networks

gchazot:
Thanks @DilaraOflaz for the deep investigation of the root cause.

The problem appeared when 2 ktd commands are run simultaneously and the dev network did
not exist prior to the commands being run. Then each command mawy create the network
separately, without the docker daemon complaining about it.

Note that this may not fix 100% of cases (both networks might end up being removed) but at least it will leave the system in a state where retrying will work.

### DIFF
--- a/kubetools/dev/backends/docker_compose/docker_util.py
+++ b/kubetools/dev/backends/docker_compose/docker_util.py
@@ -39,13 +39,21 @@ def ensure_docker_dev_network():
 
     docker_client = get_docker_client()
 
-    network_names = [
-        network.name
-        for network in docker_client.networks.list()
-    ]
-
-    if 'dev' not in network_names:
-        docker_client.networks.create('dev')
+    try:
+        new_network = docker_client.networks.create(name='dev', check_duplicate=True)
+    except docker.errors.APIError as e:
+        if e.status_code != 409:
+            raise
+    else:
+        # There's no guarantee that check_duplicate will be enforced.
+        # If we still created a duplicate, remove it
+        dev_network_ids = [
+            network.id
+            for network in docker_client.networks.list()
+            if network.name == 'dev'
+        ]
+        if len(dev_network_ids) > 1:
+            new_network.remove()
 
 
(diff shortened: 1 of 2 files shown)


kubetools/dev/backends/docker_compose/docker_util.py
```diff
@@ -39,13 +39,21 @@ def ensure_docker_dev_network():
...
+    else:
+        # There's no guarantee that check_duplicate will be enforced.
+        # If we still created a duplicate, remove it
+        dev_network_ids = [
+            network.id
+            for network in docker_client.networks.list()
+            if network.name == 'dev'
+        ]
+        if len(dev_network_ids) > 1:
+            new_network.remove()
```
DilaraOflaz, line 56: I see one small issue here.

`ktd test` create `dev` network and another unique network.
So we won't have any issue by removing the `new_network`, but what if this `dev` network is created by another job that relies on `dev` network (`ktd up`) and it contains some containers.

From what I can see in [the documentation](https://docker-py.readthedocs.io/en/stable/networks.html#docker.models.networks.Network.remove) `remove()` can raise an error.

```
remove(): Remove this network
Raises: docker.errors.APIError – If the server returns an error.
```

It won't be more safe to get the list of containers in the network and delete if `len(new_network.containers) == 0`

Or something like:
```
if len(dev_network_ids) > 1:
    for id in dev_network_ids:
        network = docker_client.networks.get(id)
        if len(network.containers) == 0:
            network.remove()
            break;
```

gchazot: We don't want to be touching any container here, only the problem network.

* `new_network` is the `dev` network that was just created at line 43 above. It is very unlikely that any other `ktd` process will have created any container in that network in the meantime.
* We don't want to be deleting any `dev` network created by any other process. They will have the same logic and they will delete their own network if needed.

DilaraOflaz (approved):

gchazot:
Thanks @DilaraOflaz

PR #128, opened 2022-06-20
Add classifiers to setup.py

gchazot:
This will also resolve the broken badge in the readme that relies on these classifiers


ryan109 (approved):

gchazot:
Thanks @ryan109

Issue #130, opened 2022-06-23
[BUG] - Kubetools v13.9.0 doesn't support Kubernetes versions prior to v1.21


DilaraOflaz:
**Describe the bug**
Kubetools `v13.9.0` doesn't support Kubernetes versions prior to `v1.21+`

**To Reproduce**
If you have a Kubernetes version prior to `v1.21` installed, here is how to reproduce the behavior:
- Run `kubetools deploy ${NAMESPACE}`

**Expected behavior**
```
kubernetes.client.exceptions.ApiException: (404)
Reason: Not Found
HTTP response headers: HTTPHeaderDict({'Server': 'openresty/1.15.8.1', 'Date': 'Wed, 22 Jun 2022 11:13:02 GMT', 'Content-Type': 'application/json', 'Content-Length': '174', 'Connection': 'keep-alive', 'Audit-Id': '', 'X-Content-Type-Options': 'nosniff', 'Strict-Transport-Security': 'max-age=15724800; includeSubDomains'})
HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"the server could not find the requested resource","reason":"NotFound","details":{},"code":404}
```

**Additional context**
CronJob `apiVersion: batch/v1` is supported by Kubernetes version `v1.21+`, [here is the doc](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/)
For lower versions we have to use `apiVersion: batch/v1beta1`
The latest version that supports `apiVersion: batch/v1beta1` is `v1.24+`, [here is the doc](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#cronjob-v125)

Issue #129, opened 2022-06-22
[BUG] - Python Kubernetes library doesn't support CronJobs


DilaraOflaz:
**Describe the bug**
Kubernetes library versions prior to `21.7.0` don't support CronJobs.

**To Reproduce**
1. In `kubetools/setup.py` modify line 59 to:
```
    'kubernetes==20.13.0',
```
2. Run `pip install -e .[dev]`
3. Run `kubetools deploy ${NAMESPACE}`. 

**Expected behavior**
`AttributeError: 'BatchV1Api' object has no attribute 'list_namespaced_cron_job'`

PR #131, opened 2022-06-23
Fix Kubernetes library version and cronjob ApiVersion

DilaraOflaz:
## Purpose of PR
This PR fixes this 2 bugs:
- https://github.com/EDITD/kubetools/issues/129
- https://github.com/EDITD/kubetools/issues/130 

Only kubetools commands and only for CronJobs


kubetools/kubernetes/api.py
```diff
@@ -41,9 +41,29 @@ def _get_k8s_apps_api(env):
...
+
+    api_core = _get_k8s_core_api(env)
+    list_node = api_core.list_node().items
+    first_node_info = list_node[0].status.node_info
+    kubelet_version = first_node_info.kubelet_version.replace('v', '').split(".")
+    major_version = int(kubelet_version[0])
+    minor_version = int(kubelet_version[1])
+
+    if major_version > 1 or (major_version == 1 and minor_version >= 21):
+        settings.IS_CRONJOB_COMPATIBLE = True
```
gchazot: settings should be constants throughout the lifetime of the program. If computation is required, then it should be done in settings itself, at the initialisation of the software, and not depending on user input.

Basically, what we're trying to do here is not a setting.

Instead it's a dynamic check to use a version or another of the API, depending on the target cluster (which is a user input)

setup.py
```diff
@@ -56,7 +56,7 @@ def get_readme_content():
             'requests>=2,<3',
             'pyretry',
             'setuptools',
-            'kubernetes',
+            'kubernetes>=21.7.0',
```
gchazot: This should be in a separate commit as it fixes a separate problem

kubetools/kubernetes/api.py
```diff
@@ -41,9 +41,29 @@ def _get_k8s_apps_api(env):
...
+
+
+def _get_k8s_batch_api(env, cronjob=False):
     api_client = _get_api_client(env)
-    return client.BatchV1Api(api_client=api_client)
+
+    if cronjob is False or check_if_cronjob_compatible(env):
+        return client.BatchV1Api(api_client=api_client)
+    else:
+        return client.BatchV1beta1Api(api_client=api_client)
```
gchazot: I think that the decision should be to use the V1 API **if compatible**, otherwise use the Beta API.

This should not depend on whether we want to create a CronJob or not, only whether the target cluster is compatible

kubetools/kubernetes/api.py
```diff
@@ -41,9 +41,29 @@ def _get_k8s_apps_api(env):
     return client.AppsV1Api(api_client=api_client)
 
 
-def _get_k8s_batch_api(env):
+def check_if_cronjob_compatible(env):
+    settings = get_settings()
+
+    api_core = _get_k8s_core_api(env)
+    list_node = api_core.list_node().items
+    first_node_info = list_node[0].status.node_info
```
gchazot: we shouldn't take just any node's version but the lowest one of all nodes.

kubetools/kubernetes/api.py
```diff
@@ -41,9 +41,29 @@ def _get_k8s_apps_api(env):
...
+    kubelet_version = first_node_info.kubelet_version.replace('v', '').split(".")
+    major_version = int(kubelet_version[0])
+    minor_version = int(kubelet_version[1])
```
gchazot: Parsing should be extracted in a function, and possibly use [an external library](https://pypi.org/project/version/) instead of reinventing the wheel.

gchazot (changes requested):

kubetools/kubernetes/api.py
DilaraOflaz: But Jobs and CronJobs use `BatchV1Api` API, I don't know how it will affect the Jobs CRUD operations.
I mean that is why I had to add the `cronjob=False` parameter

gchazot: I see, then we probably need to change the split of methods here, and have separate funcrtions to get the API for Jobs and CronJobs

kubetools/kubernetes/api.py
DilaraOflaz: https://github.com/EDITD/kubetools/pull/131/commits/d2de3d7065c1b7a84fe66a1c50a91d5fbd3834ab

kubetools/kubernetes/api.py
DilaraOflaz: I used [this one](https://pypi.org/project/packaging/)
https://github.com/EDITD/kubetools/pull/131/commits/d2de3d7065c1b7a84fe66a1c50a91d5fbd3834ab

kubetools/kubernetes/api.py
DilaraOflaz: https://github.com/EDITD/kubetools/pull/131/commits/60101b300c97a5d285ed3cb67bd119d022550e92

kubetools/kubernetes/api.py
```diff
@@ -41,11 +42,40 @@ def _get_k8s_apps_api(env):
...
+    # Get the lowest kubelet version of all the nodes of the Cluster
+    for node in list_node:
+        kubelet_version = node.status.node_info.kubelet_version.replace('v', '')
+        if version.parse(kubelet_version) < version.parse(lowest_version):
+            lowest_version = kubelet_version
+
+    if version.parse(lowest_version) >= version.parse(k8s_version):
+        settings.IS_CRONJOB_COMPATIBLE = True
+
+    return settings.IS_CRONJOB_COMPATIBLE
```
gchazot:
```suggestion
    return version.parse(lowest_version) >= version.parse(k8s_version):
```

Also, rename `k8s_version` into `required_k8s_version`

kubetools/kubernetes/api.py
gchazot: Please remove `IS_CRONJOB_COMPATIBLE`, this is not a setting. Instead, use the `check_if_cronjob_compatible` function (see [this suggestion](https://github.com/EDITD/kubetools/pull/131#discussion_r908344658))

kubetools/kubernetes/api.py
```diff
@@ -41,11 +42,40 @@ def _get_k8s_apps_api(env):
...
+    k8s_version = "1.21.0"
+    lowest_version = "1.21.0"
+
+    # Get the lowest kubelet version of all the nodes of the Cluster
+    for node in list_node:
+        kubelet_version = node.status.node_info.kubelet_version.replace('v', '')
+        if version.parse(kubelet_version) < version.parse(lowest_version):
+            lowest_version = kubelet_version
+
+    if version.parse(lowest_version) >= version.parse(k8s_version):
```
gchazot:
```suggestion
    k8s_version = version.parse("1.21.0")
    
    nodes_versions = [
        node.status.node_info.kubelet_version.replace('v', '')
        for node in list_node
    ]
    parsed_versions = map(version.parse, nodes_versions)
    lowest_version = min(parsed_version)
    
    if lowest_version >= k8s_version:
```

* pythonic
* readable: no need for a comment, variable and function names explain what's happening
* readable: reduction in function calls and operations
* more performant: parsing of each version is done exactly once
* safer: if there is no node, this will raise a `ValueError` in the `min` call.

kubetools/kubernetes/api.py
DilaraOflaz: This was my first approach but the issue with this, is that I don't have access to `env` variable in [here](https://github.com/EDITD/kubetools/pull/131/files?diff=unified&w=0#diff-8c80c5198a3e12b29443be1e74e33e55e5bd044c05c5d81c6ac2e5cf1acb6e2eR41-R45) and in [here](https://github.com/EDITD/kubetools/pull/131/files?diff=unified&w=0#diff-e6052a6275d2795b5c877cccbb45937c4927302414115c7e344761089433b107R25)

I am still trying to figure out how to access and pass `env` variable to `check_if_cronjob_compatible(env)`

DilaraOflaz: Here https://github.com/EDITD/kubetools/pull/131/commits/151a47065791870b477884869c2306d1589db324

kubetools/kubernetes/api.py
DilaraOflaz: Both are fixed in this commit https://github.com/EDITD/kubetools/pull/131/commits/151a47065791870b477884869c2306d1589db324

kubetools/kubernetes/api.py
gchazot: The change will have to be wider than this from what I can see.

A core problem shows up is that the configuration generated now depends on the version of the target k8s cluster, but we  don't always have a target server, for example in tests, and also in the `config` command.

1) We should make sure that we don't fail if there is no Cronjob definition in the config file. This is important for backwards compatibility
2) For unit tests, the functionality of using the `env` variable to figure out the version of the target cluster could be mocked. This also means that additional unit tests need to be added to check that functionality.
3) For the `config` command, we need to figure out what solution we want to adopt. One solution would be to add a flag to the command line to force the user to pass the target environment. This goes a little against the idea of the commad though which is to generate k8s config files independently from a cluster. Another solution would be to add something to the configuration file that we can use to determine the version of the batch API to use. I have a preference on this one, with a new entry under `cronjobs` specifying the version of the k8s API to use (`batch-api-version` ?). We could even make it default to the most recent version of the API to make it more transparent to the user.

DilaraOflaz: > We could even make it default to the most recent version of the API to make it more transparent to the user

Let's suppose that the user use k8s version < 1.21+ and doesn't provide the API version, if by default we put the new batch API version, the code will crash with an unreadable error message, I think it's better to force it, no ?

gchazot: Let's suppose that the user uses k8s version > 1.24 and doesn't provide the API version instead, this is much more likely to happen. And it will make the change transparent to anyone using k8s >= 1.21, which is even more likely to be the common case.

DilaraOflaz: Here is the fix for the config command: https://github.com/EDITD/kubetools/pull/131/commits/d79ae78dcdf17ba3e3f103caecc8bd705a594a94

I tested with and without the entry `batch-api-version` in CronJobs.
Also I tested without CronJobs in the config file.

Let me know what you think
(I am working on the test part, now)

kubetools/config.py
```diff
@@ -151,8 +151,9 @@ def _check_min_version(config):
 
 def _filter_config_data(key, items_or_object, env, namespace, dev):
     def is_match(item):
+        conditions = None if 'batch/v1' in item else item.get('conditions')
         return _conditions_match(
-            item.get('conditions'),
+            conditions,
             env=env,
             namespace=namespace,
             dev=dev,
```
gchazot: Please explain this change... it looks like a hack in a function that is completely unrelated (I think this function is handling `conditions` which are filters for which environments an entity is supposed to run in). So it looks like it is causing a side effect that (by coincidence) causes the change to look like what we want to do.

TLDR: I don't see how this change does what we want here

DilaraOflaz: What we want to have:
```
cronjobs:
  batch-api-version: 'batch/v1beta1'   # Must add if k8s version < 1.21+
  my-cronjob:
    schedule: "*/1 * * * *"
....
```

If I don't add this change, it tries to run `item.get('conditions')` on `'batch/v1beta1'` and it throws the following exception `--> Unexpected exception: AttributeError("'str' object has no attribute 'get'",)`

This function will return the filtered config, and we need `batch-api-version: ..` in it.
The function `_conditions_match` in `is_match(item)` return True or False to keep the config (or not) in the final return. So yes, it's a small hack to force the function to keep `batch-api-version: ..` in the final return so I can access this value [here](https://github.com/EDITD/kubetools/blob/d79ae78dcdf17ba3e3f103caecc8bd705a594a94/kubetools/kubernetes/config/cronjob.py#L23-L25)

Another thing that I can do, is to remove this change and modify [this part](https://github.com/EDITD/kubetools/blob/d79ae78dcdf17ba3e3f103caecc8bd705a594a94/kubetools/config.py#L168-L173):
```

    elif isinstance(items_or_object, dict):
        return {
            key: item
            for key, item in items_or_object.items()
            if 'batch/v1' in item or is_match(item)
        }
```

kubetools/kubernetes/api.py
DilaraOflaz: Here is a fix for the tests: https://github.com/EDITD/kubetools/pull/131/commits/8839b6055f2594a8df839ca96a4a8e7ced993814

The build is failing because of `flake8` command not because of the tests

kubetools/config.py
gchazot: Looks like the abstractions don't match then: `cronjobs` expects a dict of `job_name` -> `job_config`, so there is no place in there for a global setting applying to all cronjobs.

* Either we need to make a place for that config, but that would mean choosing a special key for the config, like you tried, in a place where all keys are valid and equivalent. This means a special handling of that key (with an explicit check on the key, not on the value), which is what your last suggestion is. Another place to put this config, instead of under `cronjobs`, would be at the top-level of the configuration file, ideally under a new section (`options` ? `settings` and `config` names are already taken)
* Or we reunite the abstractions by moving the config to a place where we can. So far the only place I can see is within each `cronjob`, for example:
  ```
  ...
  cronjobs:
      my-cronjob:
        batch-api-version: 'batch/v1beta1'   # Must add if k8s version < 1.21+
        schedule: "*/1 * * * *"
  ....
  ```
  Not ideal because it means repeating the setting for each cronjob. But if it should make the handling code simpler.

.travis.yml
```diff
@@ -14,7 +14,7 @@ python:
 install:
   # upgrade importlib_metadata due to bug in TravisCI
   # https://travis-ci.community/t/build-error-for-python-3-7-on-two-different-projects/12895
-  - 'pip install --upgrade importlib_metadata'
+  - 'pip install --upgrade "importlib_metadata<5"'
```
DilaraOflaz, line 17: This is apparently a known issue: https://github.com/python/importlib_metadata/issues/406

ryan109 (approved):
I've tested the following:

cronjob deploy < k8s v1.21
- with `batch-api-version: 'batch/v1beta1'`
deploys 🎉 
- with `batch-api-version: 'batch/v1'`
fails with ⛔ 
>Kubernetes version < 1.21 does not support Cronjob with"batch-api-version: batch/v1", you will need to use "batch-api-version: batch/v1beta1"
- without `batch-api-version` set
fails with ⛔ 
>Kubernetes version < 1.21 does not support Cronjob with"batch-api-version: batch/v1", you will need to use "batch-api-version: batch/v1beta1"

cronjob deploy > k8s v1.21
- with `batch-api-version: 'batch/v1beta1'`
deploys 🎉 
- with `batch-api-version: 'batch/v1'`
deploys 🎉 
- without `batch-api-version` set 
deploys 🎉 

The only edge case we found is k8s releases > 25 no longer supports the `batch/v1beta1` API, so we need to remove this functionality in future a kubetools version and unpin kubernetes.

ryan109:
https://github.com/EDITD/kubetools/issues/138

kubetools/config.py
ryan109: [This commit](https://github.com/EDITD/kubetools/pull/131/commits/1c195851a9673b011f7d865b85b5b31edcef278d) adds the api version inside individual cronjobs - I'm going to resolve this @gchazot, please reopen if you don't agree with the change

ryan109 (approved):
LGTM

kubetools/kubernetes/api.py
```diff
@@ -41,11 +45,81 @@ def _get_k8s_apps_api(env):
...
+
+
+def get_cronjob_api_version(obj):
+    if obj is not None:
+        if isinstance(obj, dict):
+            if obj['apiVersion'] is not None:
+                return obj['apiVersion']
+        else:
+            if obj.api_version is not None:
+                return obj.api_version
```
gchazot: Please use a meaningful variable name instead of `obj`

tests/configs/k8s_cronjobs_beta_api_version/kubetools.yml
```diff
@@ -0,0 +1,27 @@
...
+      generic-container:
+        containerContext: generic-context
+
+  generic-cronjob-2:
+    batch-api-version: 'batch/v1beta1'
+    schedule: "*/1 * * * *"
+    concurrency_policy: "Allow"
+    containers:
+      generic-container:
+        containerContext: generic-context
```
gchazot: Changing the values in one of the 2 cronjobs will improve the resilience of the test to some forms of bugs. It will also improve the readability of the test.
For example having different schedules, different concurrency policies etc.

gchazot (commented):
Just a couple of minor improvements to do

kubetools/kubernetes/api.py
DilaraOflaz: https://github.com/EDITD/kubetools/pull/131/commits/1ae2c19d94e357887f58d750197c8ee5892540f8

tests/configs/k8s_cronjobs_beta_api_version/kubetools.yml
DilaraOflaz: https://github.com/EDITD/kubetools/pull/131/commits/3c85287e7dc20f22bd8b17674a8daf2aa5f52969

gchazot (approved):
LGTM

PR #132, opened 2022-07-05
Job objects in kubetools yaml

ryan109:
## Related links


ryan109:
Going to rethink this, we only allow `upgrade` jobs to be passed in via `kubetools.yml` and the config creation around it is not very intuitive for if we wanted to pass job objects that adhered to a similar structure as our other k8s resources.

PR #134, opened 2022-07-13
Fix jobs resources

fmcwalters-edited:
## Purpose of PR
- Parse job resources in the job config method from job spec, instead of passing it separately

## Related links

### DIFF
--- a/kubetools/kubernetes/config/job.py
+++ b/kubetools/kubernetes/config/job.py
@@ -57,8 +57,7 @@ def make_job_config(
     )
 
     # if resources exist in job config, use it here
-    if config.get('resources'):
-        resources = config.get('resources')
+    resources = config.get('resources', {})
 
     # Make our container
     container = make_container_config(
--- a/kubetools/kubernetes/config/job.py
+++ b/kubetools/kubernetes/config/job.py
@@ -13,7 +13,6 @@ def make_job_config(
     envvars=None,
     job_name=None,
     container_name="upgrade",
-    resources=None,
 ):
     '''
     Builds a Kubernetes job configuration dict.
@@ -56,9 +55,6 @@ def make_job_config(
         {'KUBE_JOB_ID': job_id},
     )
 
-    # if resources exist in job config, use it here
-    resources = config.get('resources', {})
-
     # Make our container
     container = make_container_config(
         job_id,
@@ -67,7 +63,7 @@ def make_job_config(
             'command': command,
             'image': config['image'],
             'chdir': config.get('chdir', '/'),
-            'resources': resources,
+            'resources': config.get('resources', {}),
         },
         envvars=envvars,
         labels=labels,
--- a/kubetools/kubernetes/config/job.py
+++ b/kubetools/kubernetes/config/job.py
@@ -56,6 +56,10 @@ def make_job_config(
         {'KUBE_JOB_ID': job_id},
     )
 
+    # if resources exist in job config, use it here
+    if config.get('resources'):
+        resources = config.get('resources')
+
     # Make our container
     container = make_container_config(
         job_id,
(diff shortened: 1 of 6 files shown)


tests/test_make_job_config.py
```diff
@@ -37,3 +37,14 @@ def test_container_name_can_be_set_by_caller(self):
         job_config = make_job_config(load_job_spec(), container_name='mycoolcontainer')
         container_name = job_config['spec']['template']['spec']['containers'][0]['name']
         self.assertEqual('mycoolcontainer', container_name)
+
+    def test_resources_being_passed(self):
+        test_resources = {
+                "requests": {
+                    "cpu": 1,
```
fmcwalters-edited: I think I should delete this CPU request as we know there are errors with doing CPU resource limits and it may give end users a false sense of security

fmcwalters-edited: [8c0c5a7](https://github.com/EDITD/kubetools/pull/134/commits/8c0c5a70eb21e3209f2a3c1ec7e4bb242b4d34c2)

ryan109 (approved):

kubetools/kubernetes/config/job.py
```diff
@@ -56,6 +56,10 @@ def make_job_config(
         {'KUBE_JOB_ID': job_id},
     )
 
+    # if resources exist in job config, use it here
+    if config.get('resources'):
+        resources = config.get('resources')
```
gchazot: I believe it would do the exact same as this
```suggestion
    # if resources exist in job config, use it here
    resources = config.get('resources')
```
Then, the function argument can be removed and the variable inlined since it is only used in 1 place

tests/test_make_job_config.py
```diff
@@ -37,3 +37,13 @@ def test_container_name_can_be_set_by_caller(self):
...
+
+    def test_resources_being_passed(self):
+        test_resources = {
+                "requests": {
+                    "memory": "1Gi",
+                },
+        }
+        job_config = make_job_config(load_job_spec())
+        resource_config = job_config['spec']['template']['spec']['containers'][0]['resources']
+        self.assertEqual(resource_config, test_resources)
```
gchazot: * Naming
* By convention, expected comes first in assertions
```suggestion
        expected_resources = {
                "requests": {
                    "memory": "1Gi",
                },
        }
        job_config = make_job_config(load_job_spec())
        resource_config = job_config['spec']['template']['spec']['containers'][0]['resources']
        self.assertEqual(expected_resources, resource_config)
```

gchazot (changes requested):
Couple of changes needed.

kubetools/kubernetes/config/job.py
ryan109: [9387f5d](https://github.com/EDITD/kubetools/pull/134/commits/9387f5d3f7d1b716aa1c60e3145846c33f398e77)

ryan109: Think empty dict is what we want to return here over `None`? Not sure `None` fits the k8s job spec as well

gchazot: `None` is (was?) the default value of the function argument

gchazot: > Then, the function argument can be removed and the variable inlined since it is only used in 1 place

Still pending

ryan109: Yeh, `None` could work but I think `{}` is more inline with what I would expect to see if I saw a job_config with no resources defined.

Done the second part now!
[fe17c2e](https://github.com/EDITD/kubetools/pull/134/commits/fe17c2e0ca4984f8bee291b52669db87f93679dc)

gchazot (approved):
Curious how this feels like this implements the [comment from my previous review](https://github.com/EDITD/kubetools/pull/126#pullrequestreview-1008637692)

ryan109:
Thanks @gchazot, yeh looking back we should have just done it this way. Not sure what I was thinking when I replied...

PR #135, opened 2022-12-19
Add support for job time to live

balbirthomas:
Failure to remove resources used by completed jobs/pods can lead to resource exhaustion in the Kubernetes cluster. This commit adds support for setting a ttlSecondsAfterFinished parameter in the Kubernetes Job sepc. This parameter specifies the amount of time (in seconds) for which the resources used by completed job/pod will persist after it which it will be garbage collected by the Kube control manager.

The newly added parameter is optional and only set if it passed in through the `config` argument to the `make_job_spec` method.

## Related links

### DIFF
--- a/kubetools/kubernetes/config/job.py
+++ b/kubetools/kubernetes/config/job.py
@@ -75,7 +75,7 @@ def make_job_config(
     # Parallelism defaults to completions, also as Kubernetes
     parallelism = config.get('parallelism', completions)
 
-    return {
+    job_config = {
         # Normal Kubernetes job config
         'apiVersion': 'batch/v1',
         'kind': 'Job',
@@ -99,3 +99,8 @@ def make_job_config(
             },
         },
     }
+
+    if 'ttl_seconds_after_finished' in config:
+        job_config['spec']['ttlSecondsAfterFinished'] = config.get('ttl_seconds_after_finished')
+
+    return job_config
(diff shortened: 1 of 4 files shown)


kubetools/kubernetes/config/job.py
```diff
@@ -99,3 +99,8 @@ def make_job_config(
             },
         },
     }
+
+    if config.get('ttlSecondsAfterFinished'):
```
gchazot: `if 'ttlSecondsAfterFinished' in config`:

gchazot (commented):
Please check if tests can be added for this new flag, and if any documentation/samples needs updated

kubetools/kubernetes/config/job.py
balbirthomas: Ack. Will change.

balbirthomas:
> Please check if tests can be added for this new flag, and if any documentation/samples needs updated

Ack. Will do and ping you for review.

kubetools/kubernetes/config/job.py
balbirthomas: Changes made and set for review.

gchazot (approved):
LGTM.
Please be aware that, even when this is released into a new version of kubetools, it will still [not work on clusters running k8s < 1.21](https://github.com/EDITD/kubetools/issues/130).

One solution may be to backport this onto a branch based on `kubetools 13.8.1` and release that as a 13.8.x version

seangarner (approved):

PR #136, opened 2022-12-20
add timeoutSeconds for `probes` that use `exec`

fmcwalters-edited:
as of Kubernetes 1.20, the `timeoutSeconds` is now enforced, and has a default value of 1 second. Add this field in and increase to 5 seconds to match other `timeoutSeconds` defined within Kubetools

PFM-250

## Related links


gchazot (changes requested):
I don't think these ones ever get loaded on any `k8s` cluster, but only used to test the generated files

gchazot:
Closing as this is not necessary

PR #141, opened 2023-01-26
Update optional default registry to be command line argument

DilaraOflaz:
## Purpose of PR
Updated kubetools to use the default_registry argument as the prefix of images specified in `kubetools.yml` file, rather than expecting the registry name to be in the image name

### DIFF
--- a/kubetools/kubernetes/config/__init__.py
+++ b/kubetools/kubernetes/config/__init__.py
@@ -35,8 +35,19 @@ def _ensure_image(
     container, context_name_to_image,
     deployment_name=None,
     container_name=None,
+    default_registry=None,
 ):
     if 'image' in container:
+        registry_image = container['image'].split('/')
+        if len(registry_image) == 1:
+            # If the image is commonly available from DockerHub
+            if default_registry is None:
+                container['image'] = registry_image[0]
+
+            # If registry not specified in the kubetools.yml file
+            else:
+                image = container['image']
+                container['image'] = f'{default_registry}/{image}'
         return
 
     if 'containerContext' in container:
@@ -53,7 +64,12 @@ def _ensure_image(
     container['image'] = context_name_to_image[context_name]
 
 
-def _get_containers_data(containers, context_name_to_image, deployment_name):
+def _get_containers_data(
+    containers,
+    context_name_to_image,
+    deployment_name,
+    default_registry=None,
+):
     # Setup top level app service mapping all ports from all top level containers
     all_container_ports = []
     all_containers = {}
@@ -63,6 +79,7 @@ def _get_containers_data(containers, context_name_to_image, deployment_name):
         _ensure_image(
             data, context_name_to_image,
             deployment_name, container_name,
+            default_registry=default_registry,
         )
         all_containers[container_name] = data
 
@@ -119,6 +136,7 @@ def generate_kubernetes_configs_for_project(
     # Map of context name -> docker image - images are expected to be built by
     # something else (normally the kubetools server or ktd client).
     context_name_to_image=None,
+    default_registry=None,
 ):
     '''
     Builds service & deployment definitions based on the app config provided. We have
@@ -163,6 +181,7 @@ def generate_kubernetes_configs_for_project(
             dependency['containers'],
             context_name_to_image=context_name_to_image,
             deployment_name=name,
+            default_registry=default_registry,
         )
         app_annotations = copy_and_update(base_annotations)
 
@@ -196,6 +215,7 @@ def generate_kubernetes_configs_for_project(
             deployment['containers'],
             context_name_to_image=context_name_to_image,
             deployment_name=name,
+            default_registry=default_registry,
         )
         app_annotations = copy_and_update(
             base_annotations,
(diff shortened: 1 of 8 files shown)


kubetools/kubernetes/config/__init__.py
```diff
@@ -35,8 +35,19 @@ def _ensure_image(
...
+        registry_image = container['image'].split('/')
+        if len(registry_image) == 1:
+            # If the image is commonly available from DockerHub
+            if default_registry is None:
+                container['image'] = registry_image[0]
+
+            # If registry not specified in the kubetools.yml file
+            else:
+                image = container['image']
+                container['image'] = f'{default_registry}/{image}'
```
gchazot:
```suggestion
        if len(registry_image) == 1 and default_registry is not None:
            # If registry is not specified but a default was provided
            container['image'] = f'{default_registry}/{container['image']}'
```

The `if` part of the `if..else` was a no-op

gchazot (commented):
Please add some test to expose the functionality

kubetools/kubernetes/config/__init__.py
DilaraOflaz: https://github.com/EDITD/kubetools/pull/141/commits/c7856312ccb204374777d69ef580e151d7be594c

DilaraOflaz:
I added tests in [this commit](https://github.com/EDITD/kubetools/pull/141/commits/cf48a44910ede470020b35cb75db55ead9040619)

tests/configs/docker_registry/k8s_services.yml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+kind: Service
+metadata:
+  annotations: {app.kubernetes.io/managed-by: kubetools}
+  labels: {kubetools/name: generic-app, kubetools/project_name: generic-app, kubetools/role: app}
+  name: generic-app
+spec:
+  ports:
+  - {port: 80, targetPort: 80}
+  - {port: 80, targetPort: 80}
```
gchazot:
```suggestion
  - {port: 80, targetPort: 80}
```

tests/configs/docker_registry/kubetools.yml
```diff
@@ -0,0 +1,24 @@
...
+    containers:
+      with-default-registry:
+        command: [uwsgi, --ini, /etc/uwsgi.conf]
+        containerContext: generic-context
+
+      without-default-registry:
+        image: generic-image
+        command: [uwsgi, --ini, /etc/uwsgi.conf]
+        ports:
+          - 80
```
gchazot: Needs some renaming here:
* `with-default-registry` has a non-default registry in the containerContext
* `without-default-registry` doesn't have a registry, hence uses the default registry

Also worth changing the test to avoid mixing 2 things: one is using a containerContext, the other is not. This is not what we're testing here. So either there needs to be 4 cases (with/without registry/context) or only 2 that are identical except for the with/without registry aspect

tests/configs/docker_registry/k8s_deployments.yml
```diff
@@ -0,0 +1,30 @@
...
+  selector:
+    matchLabels: {kubetools/name: generic-app, kubetools/project_name: generic-app,
+      kubetools/role: app}
+  template:
+    metadata:
+      labels: {kubetools/name: generic-app, kubetools/project_name: generic-app, kubetools/role: app}
+    spec:
+      containers:
+      - command: [generic-command]
+        containerContext: generic-context
```
gchazot: Is this supposed to be in there?

tests/test_config_generation.py
```diff
@@ -60,3 +61,6 @@ def test_k8s_cronjobs_beta_api_version_configs(self):
 
     def test_multiple_deployments_configs(self):
         _test_configs('multiple_deployments')
+
+    def test_registry_configs(self):
+        _test_configs('docker_registry', 'default-registry')
```
gchazot: Why is this not following the same convention as all the other tests, named `test_<folder_name>_configs`?

gchazot:
```suggestion
    def test_registry_configs(self):
        _test_configs('docker_registry', default_registry='default-registry')
```
Explicit > implicit

gchazot (changes requested):

tests/configs/docker_registry/k8s_services.yml
DilaraOflaz: https://github.com/EDITD/kubetools/pull/141/commits/b69af2413f6bf45f793c6462ce127b2478927dd2

tests/configs/docker_registry/kubetools.yml
DilaraOflaz: https://github.com/EDITD/kubetools/pull/141/commits/b69af2413f6bf45f793c6462ce127b2478927dd2

tests/configs/docker_registry/k8s_deployments.yml
DilaraOflaz: https://github.com/EDITD/kubetools/pull/141/commits/b69af2413f6bf45f793c6462ce127b2478927dd2

tests/test_config_generation.py
DilaraOflaz: https://github.com/EDITD/kubetools/pull/141/commits/b69af2413f6bf45f793c6462ce127b2478927dd2

DilaraOflaz: https://github.com/EDITD/kubetools/pull/141/commits/b69af2413f6bf45f793c6462ce127b2478927dd2

gchazot (approved):
LGTM

DilaraOflaz:
Thanks @gchazot

PR #142, opened 2023-02-06
enable_ipv6_networks - Enabling support for IPv6

svedbg:
## Purpose of PR
At the moment IPv6 is not enabled in dev networks, so containers don't get ip address assigned. With this PR IPv6 is allowed on the network and is up to application if ipv4 or ipv6 will be used.

Network

### DIFF
--- a/kubetools/dev/backends/docker_compose/docker_util.py
+++ b/kubetools/dev/backends/docker_compose/docker_util.py
@@ -40,7 +40,11 @@ def ensure_docker_dev_network():
     docker_client = get_docker_client()
 
     try:
-        new_network = docker_client.networks.create(name='dev', check_duplicate=True)
+        new_network = docker_client.networks.create(
+            name='dev',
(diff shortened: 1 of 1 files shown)


ryan109 (commented):
We will need to add a changelog and release notes.

Has this been tested locally? What effect does enabling ipv6 doesn't have on ipv4, if any?

svedbg:
@ryan109 the ipv6 was tested locally and is working, the issue is that it's not supported on windows/mac. On top of that we have issues with running tests, because they create separate network for each container. I think docker support of ipv6 is not ready to be used in production: https://docs.docker.com/config/daemon/ipv6/

ryan109:
I don't think we can merge this then, it's likely that attempting to create an ipv6 network on macos wouldn't work so we would break kubetools for all mac users ⚠️ We will need to find some time to test this on a mac...I'll do it myself when I get the chance.

PR #146, opened 2023-03-06
Correct image registry handling

gchazot:
## Purpose of PR
This fixes a problem introduced with https://github.com/EDITD/kubetools/pull/141 where images that are specified directly, without a registry (hence should be assumed to be sourced from the docker daemon default registry, usually dockerhub) got modified into using the value from `--default-registry` option. That option is instead intended to point at the registry where to push the images built by kubetools and hence where k8s can pull those images from.


DilaraOflaz (approved):

tests/configs/docker_registry/Dockerfile-direct
```diff
@@ -0,0 +1,2 @@
+FROM python:3.9-slim
+ENTRYPOINT ["python"]
```
ryan109: Why do we need 2 test configs that are identical? The other being `Dockerfile-context`

ryan109 (approved):
Changes LGTM

tests/configs/docker_registry/Dockerfile-direct
gchazot: It's just to make sure they are distinguishable in the generated k8s manifests

gchazot: Actually you're right here, simplifying: 1d23331

gchazot:
Thanks for the reviews @DilaraOflaz @ryan109

PR #147, opened 2023-04-04
AWS EKS support

tmwoodruff:
## Purpose of PR

Updates to support deployment to AWS EKS.

I've run into two issues deploying to EKS:

1. ECR requires authentication and SSL, but the check for existing docker images in `deploy/image.py` makes an http request with no auth. To fix this, I've added the ability to configure an external script that can be called to check for existing images (see example below). Since there's no standard way to check registries, this provides flexibility to support whatever is needed.
3. The version compatibility check in `kubernetes/api.py:check_if_cronjob_batch_v1_compatible()` fails with version strings returned by EKS (e.g., `v1.25.7-eks-a59e1f0`). I've added a regex to strip off anything after `-` in the version string,

Kubernetes deployment

## Additional information

Example image check script that uses aws-cli:

```sh
#!/bin/bash

AWS_REGION=us-west-2

REGISTRY="$1"
APP="$2"
COMMIT_INFO="$3"

[[ "$REGISTRY" != *.dkr.ecr.*.amazonaws.com ]] && exit 2

registry_id=${REGISTRY%%.*}

echo "Checking for image: <${registry_id}> <${APP}> <${COMMIT_INFO}>"

res=`aws ecr describe-images --registry-id ${registry_id} --region ${AWS_REGION} --repository-name ${APP} --image-ids imageTag=${COMMIT_INFO} 2>&1`

[[ "$res" = *imageDetails* ]] && { echo "Found"; exit 0; }

[[ "$res" = *ImageNotFoundException* ]] && { echo "Not Found"; exit 1; }

echo "ERROR:"
echo "$res"

exit 255
```


kubetools/settings.py
```diff
@@ -23,6 +23,8 @@ class KubetoolsSettings(object):
 
     CRONJOBS_BATCH_API_VERSION = 'batch/v1'  # if k8s version < 1.21+ should be 'batch/v1beta1'
 
+    REGISTRY_CHECK_SCRIPT = None
```
Jberrz, line 20: would this be easier to set as an ENV VAR? or is it trivial to override kubetools settings as needed?

tmwoodruff: They're loaded from ` ~/.kubetools/kubetools.conf `.

Jberrz (approved):

kubetools/deploy/image.py
```diff
@@ -38,6 +41,25 @@ def has_app_commit_image(registry, app_name, context_name, commit_hash):
         raise KubeBuildError(f'Invalid registry to build {context_name}: {registry}')
 
     commit_version = get_commit_hash_tag(context_name, commit_hash)
+
+    settings = get_settings()
+    if (settings.REGISTRY_CHECK_SCRIPT):
```
gchazot:
```suggestion
    if settings.REGISTRY_CHECK_SCRIPT:
```

kubetools/kubernetes/api.py
```diff
@@ -51,8 +52,9 @@ def check_if_cronjob_batch_v1_compatible(env, batch_api_version):
     required_k8s_version = version.parse("1.21.0")
     supported_beta_version = version.parse("1.25.0")
 
+    version_pattern = re.compile(r"^v?([^-]+).*$")
     nodes_versions = [
-        node.status.node_info.kubelet_version.replace('v', '')
+        version_pattern.sub(r"\1", node.status.node_info.kubelet_version)
         for node in list_node
     ]
     parsed_versions = map(version.parse, nodes_versions)
```
gchazot: Maybe this whole function would work better rewritten using something like https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/CoreApi.md#get_api_versions. i.e. look at the API versions supported by the cluster rather than rely on the version of k8s

tmwoodruff: I don't have older cluster versions to test this on, but here's an attempt: https://github.com/EDITD/kubetools/pull/147/commits/2da736b2377bfe1dfe2214829f12b31e66eeb5b4

gchazot: I've added a fair bit of refactoring and cleanup. I'll try to test that with various k8s cluster versions

gchazot: Well. I tested and it didn't go well.
The work on "CronJobs" and version compatibility, as well as all the other bugs I found along the way are getting out of hand. I'm gonna split all of this in a couple of separate branches

gchazot: I'm going to keep only the "EKS" support in this PR and will move the rest out

gchazot: First split out: Some cleanup and bug fixes I found while testing. https://github.com/EDITD/kubetools/pull/149

gchazot: Finally, also split out the CronJob / APIVersion refactoring into https://github.com/EDITD/kubetools/pull/150

gchazot (approved):
Happy with it now. I let @tmwoodruff check it again

README.md
```diff
@@ -88,6 +88,16 @@ kubetools deploy my-namespace
 pip install kubetools
 ```
 
+## Configuration
+Users can  configure some aspects of `kubetools`. The configuration folder location depends on the
```
tmwoodruff, line 92: Extra space before 'configure'

tmwoodruff:
Looks great. Thanks, @gchazot .

gchazot:
Thanks @tmwoodruff @Jberrz

README.md
gchazot: f079741
Added it in the release PR

PR #149, opened 2023-04-11
Small fixes and cleanups

gchazot:
## Purpose of PR
* Small cleanups as seen during my work on #147 
* Fix issues with k8s namespace handling:
  * the "default" namespace cannot be deleted, that's forbidden by k8s
  * don't try to delete the target namespace if it doesn't even exist
* Fix a bug with gathering annotations when some objects in the namespace may not have any. This happens only if the kubetools application is deployed to a namespace shared with resources deployed with other tools, and that may not have any annotations, or even be of a Kind that doesn't allow annotations.

### DIFF
--- a/kubetools/deploy/commands/cleanup.py
+++ b/kubetools/deploy/commands/cleanup.py
@@ -75,7 +75,7 @@ def get_cleanup_objects(build, cleanup_jobs):
         replica_set_names - replica_set_names_to_delete - replica_set_names_already_deleted
     )
 
-    if len(remaining_pods) == 0 and len(remaining_replicasets) == 0:
+    if current_namespace and len(remaining_pods) == 0 and len(remaining_replicasets) == 0:
         namespace_to_delete = [current_namespace]
 
     return namespace_to_delete, replica_sets_to_delete, pods_to_delete, jobs_to_delete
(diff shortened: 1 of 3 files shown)


tmwoodruff (approved):
Nice improvements. Looks good to me.

kubetools/deploy/commands/cleanup.py
```diff
@@ -64,19 +64,23 @@ def get_cleanup_objects(build, cleanup_jobs):
             pod_names_already_deleted.add(get_object_name(pod))
 
     namespaces = list_namespaces(build.env)
-    current_namespace = None
     for namespace in namespaces:
         if namespace.metadata.name == build.namespace:
             current_namespace = namespace
+            break
+    else:
+        current_namespace = None
```
DilaraOflaz, line 72: I think there is an indentation issue here

gchazot: [nope](https://docs.python.org/3/tutorial/controlflow.html#break-and-continue-statements-and-else-clauses-on-loops)

DilaraOflaz: Thanks

DilaraOflaz (approved):
LGTM

gchazot:
Thanks for the reviews @DilaraOflaz @tmwoodruff

PR #150, opened 2023-04-13
Fix cronjobs and api versions

gchazot:
This is extracted and fixed/improved from #147 .
It's eventually a massive re-write of the logic to determine which API Version to use for CronJobs:
* Delaying the decision until the k8s cluster is contacted, unless the user specified an explicit version to use
* Using the information provided by the k8s APIs on a per-resource-kind level (the "API Groups" API doesn't give us the information we need here)

Eventually, we should drop support specifically for this case in `kubetools`, but this demonstrates how we could do it for any of the other k8s APIs in order to support multiple versions. It also helps understand the gaps in the code architecture that need to be filled in order to support this for other Resource kinds.

I've also fixed an important bug along the way: We shouldn't wait for cronjobs to finish executing on `deploy`; the next schedule might be a year from now. So we only need to wait for the CronJob resource to be created/updated

Sorry I don't find an easy way to clean-up the commit history better, so it still shows the numerous refactoring steps that were needed to clarify/simplify the code, and attempts that didn't work, until the solution became possible.


kubetools/kubernetes/api.py
```diff
@@ -466,3 +309,100 @@ def check_job():
...
+    possible_versions = [
+        'batch/v1',
+        'batch/v1beta1',
+    ]
+    if default_version:
+        possible_versions.insert(0, default_version)
+
+    for version in possible_versions:
+        batch_api = _get_k8s_cronjobs_batch_api(env, version)
+        api_resources = batch_api.get_api_resources()
```
tmwoodruff, line 375: Assume you've tested this, but does this raise an error when the cluster does not support a version. For example, when using an older k8s version that doesn't support `batch/v1`, does `get_api_resources()` return an empty list or raise an exception?

tmwoodruff (approved):
One small question, but otherwise looks good. Thanks for doing this!

kubetools/kubernetes/api.py
gchazot: I've tested this with k8s versions down to 1.18 and encountered no issue

gchazot:
Thanks for the review @tmwoodruff

kubetools/kubernetes/api.py
gchazot: 100644
Added that in the release PR

PR #154, opened 2023-06-27
Add build vars

gchazot:
Adding system to provide "build vars", i.e. values to pass to ARGs in the Dockerfile.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,27 @@ Install the package in editable mode, with the dev extras:
 pip install -e .[dev]
 ```
 
+## Local deployment testing
+
+For deployment testing, you will need a kubernetes cluster and a docker registry. You can get both
+easily using `minikube`:
+```shell
+minikube start --addons registry --insecure-registry ${MINIKUBE_IP}:5000
+```
+Then you can deploy to that environment:
+```shell
+kubetools --context minikube deploy --default-registry ${MINIKUBE_IP}:5000 default .
+```
+
+`MINIKUBE_IP` value can vary depending on your local environment. The easiest way to get the correct
+value is to start minikube once then reset it:
+```shell
+minikube start
+MINIKUBE_IP=$(minikube ip)
+minikube delete
+...
+```
+
 ## Releasing (admins/maintainers only)
 * Update [CHANGELOG](CHANGELOG.md) to add new version and document it
 * In GitHub, create a new release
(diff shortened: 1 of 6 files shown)


tmwoodruff (commented):
Looks good. And the code that builds the docker images is lot nicer now.

tmwoodruff (approved):

README.md
```diff
@@ -122,3 +143,15 @@ pip install -e .[dev]
...
+`kubetools` now supports passing values for `ARG` parameters used in Dockerfiles, using
+`--build-args`. This has a couple of caveats though:
+* it is NOT supported in `ktd`. A workaround for this is to use the default value of the `ARG`
+  instruction.
+* this doesn't affect the image tag pushed to the docker registry, which is based only on the git
+  commit hash. This means that these arguments cannot be used to generate multiple images from the
+  same Dockerfile. So their main usage should be to pass secrets that should not be recorded in the
+  git repository but are needed at build time, to access external resources for example.
+* these values could be recorded in the docker image layer history. To prevent leaking secrets, you
+  should consider using multi-stage builds where the secrets are only used in a "builder" image.
```
ryan109, line 157: Just to clarify on this, it _could_ effect the image pushed to the docker registry if used incorrectly.

The correct usage is to use a builder stage so when the builder is discarded/not pushed the secrets die with it, however if someone injected it into main final image this would get leaked as mentioned. The first line in the description confused me when it said it wouldn't effect the image tag pushed.

ryan109 (approved):

README.md
gchazot: I don't understand if anything is wrong in this documentation or what I could change to make anything clearer.

ryan109: I think it's just the first line, why is it necessary?
>this doesn't affect the image tag pushed to the docker registry, which is based only on the git
  commit hash.
  
  Do you mean that adding vars to the docker build won't effect the naming of the tag? In which case I don't think it would ever be implied that it would, so we shouldn't need to mention it at all.

gchazot: Because if you try to deploy twice with only different values for the build-args, since the `tag` is not affected, `kubetools` will think the image is already on the target registry (because an image with this tag was already pushed). So it won´t even try to rebuild the image

gchazot: This behaviour is necessary to keep the current image tagging system (based on git commit hash) and without risking to break the principle of immutability of a docker image tag

ryan109: Ahhh, ok now it makes sense.

ryan109 (approved):

gchazot:
Thanks @tmwoodruff @ryan109

PR #159, opened 2023-12-01
Fix dependencies

gchazot:
This fixes 2 separate problems that showed up with releases within dependencies:
* Cython 3 release broke PyYaml installation. This can be worked around by adding a constraint for Cython version upon install.
* Urllib3 release 2.0 broke the `requests` package used by `docker-py`. Since we also depend directly on `requests, we work this around by pinning it to a compatible version.

Note that these are work-around for the current state of the tool. They should be removed once a proper upgrading of dependencies is done.

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -19,8 +19,11 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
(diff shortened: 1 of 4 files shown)


tmwoodruff (approved):
Looks great. Thanks, @gchazot!

gchazot:
Thanks @tmwoodruff

PR #162, opened 2024-02-02
Use kube-context for kubetools config command

gchazot:
This ensures that the generated config matches the configs used when using the corresponding deploy command


ryan109 (approved):

gchazot:
Thanks @ryan109

PR #163, opened 2024-02-15
use docker compose v2

fmcwalters-edited:
docker compose v2 is part of Docker nowadays. Rather than calling a separate Python lib that calls `docker-compose`, call `docker compose` and use v2.
Fix the `-` and `_` difference between v1 and v2.
This will need THOROUGH testing

PFM-827

## Purpose of PR
Upgrade to use Docker Compose V2
docker compose
## Todos
Test on Linux machines, different CI builds with different packages
## Additional information
This will require Docker with Compose v2
## Related links

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 # v14.0.0-dev
 - **DO NOT USE WITH DOCKER DESKTOP v4.27.[0..2]** as this has a bug using `insecure-registries` with `docker compose` - [issue here](https://github.com/docker/buildx/issues/2030)
+- this is a `dev` release because of this bug - there is a workaround for this but we don't want to use workarounds in production
 - this **BREAKS COMPATIBILITY** with previous versions of `ktd` (see README)
 - uses `docker compose` in place of `docker-compose`, meaning we use Docker compose v2 on the backend
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,9 @@
 
 ### Unreleased
 
-# v20.0.0
-- use `docker compose` in place of `docker-compose`, meaning we use Docker compose v2 on the backend
+# v14.0.0
+- this **BREAKS COMPATIBILITY** with previous versions of `ktd` (see README)
+- uses `docker compose` in place of `docker-compose`, meaning we use Docker compose v2 on the backend
 
 # v13.13.1
 - Add nodeSelector config to kubetools file
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### Unreleased
 
-# v14.0.0-dev
+# v14.0.0
 - before upgrading to version 14.0 or above, you _must_ run `ktd destroy` for all existing 
 local Kubetools projects.
 - this is a `dev` release because of a bug in `docker/buildx` - there is a workaround for this but we don't want to use workarounds in production (see TROUBLESHOOTING in README)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,9 @@
 ### Unreleased
 
 # v14.0.0-dev
-- **DO NOT USE WITH DOCKER DESKTOP v4.27.[0..2]** as this has a bug using `insecure-registries` with `docker compose` - [issue here](https://github.com/docker/buildx/issues/2030)
-- this is a `dev` release because of this bug - there is a workaround for this but we don't want to use workarounds in production
+- before upgrading to version 14.0 or above, you _must_ run `ktd destroy` for all existing 
+local Kubetools projects.
+- this is a `dev` release because of a bug in `docker/buildx` - there is a workaround for this but we don't want to use workarounds in production (see TROUBLESHOOTING in README)
 - this **BREAKS COMPATIBILITY** with previous versions of `ktd` (see README)
 - uses `docker compose` in place of `docker-compose`, meaning we use Docker compose v2 on the backend
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### Unreleased
 
-# v14.0.0
+# v14.0.0-dev
 - **DO NOT USE WITH DOCKER DESKTOP v4.27.[0..2]** as this has a bug using `insecure-registries` with `docker compose` - [issue here](https://github.com/docker/buildx/issues/2030)
 - this **BREAKS COMPATIBILITY** with previous versions of `ktd` (see README)
 - uses `docker compose` in place of `docker-compose`, meaning we use Docker compose v2 on the backend
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Unreleased
 
 # v14.0.0
+- **DO NOT USE WITH DOCKER DESKTOP v4.27.[0..2]** as this has a bug using `insecure-registries` with `docker compose` - [issue here](https://github.com/docker/buildx/issues/2030)
 - this **BREAKS COMPATIBILITY** with previous versions of `ktd` (see README)
 - uses `docker compose` in place of `docker-compose`, meaning we use Docker compose v2 on the backend
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ### Unreleased
 
+# v20.0.0
+- use `docker compose` in place of `docker-compose`, meaning we use Docker compose v2 on the backend
+
 # v13.13.1
 - Add nodeSelector config to kubetools file
 - Fix bug where `config` command was not printing the actual `k8s` configs used by `deploy` because it did not take into account the kube-context, whether default or given with `--context`.
--- a/README.md
+++ b/README.md
@@ -85,11 +85,13 @@ ktd up
 ```
 
 **NOTE**: there is a bug in the Docker BuildKit that ignores "insecure-registries". 
-as `docker compose` V2 uses BuildKit by default, the only way to currently get around this 
+As `docker compose` V2 uses BuildKit by default, the only way to currently get around this 
 is to append `DOCKER_BUILDKIT=0` in front of the `ktd` command,
+
 e.g: `DOCKER_BUILDKIT=0 ktd up`.
+
 This applies to all `Dockerfile`s where an insecure registry is specified.
- 
+
 ```sh
 # Deploy the project to a Kubernetes namespace
 kubetools deploy my-namespace
--- a/README.md
+++ b/README.md
@@ -82,12 +82,23 @@ With this in your current directory, you can now:
 ```sh
 # Bring up a local development environment using docker compose
 ktd up
+```
 
+**NOTE**: there is a bug in the Docker BuildKit that ignores "insecure-registries". 
+as `docker compose` V2 uses BuildKit by default, the only way to currently get around this 
+is to append `DOCKER_BUILDKIT=0` in front of the `ktd` command,
+e.g: `DOCKER_BUILDKIT=0 ktd up`.
+This applies to all `Dockerfile`s where an insecure registry is specified.
+ 
+```sh
 # Deploy the project to a Kubernetes namespace
 kubetools deploy my-namespace
 ```
 
 ## Installing
+
+**NOTE**: before upgrading to version 14.0 or above, you _must_ run `ktd destroy` for all existing 
+local Kubetools projects.
 
 ```sh
 pip install kubetools
--- a/README.md
+++ b/README.md
@@ -84,14 +84,6 @@ With this in your current directory, you can now:
 ktd up
 ```
 
-**NOTE**: there is a bug in the Docker BuildKit that ignores "insecure-registries". 
-As `docker compose` V2 uses BuildKit by default, the only way to currently get around this 
-is to append `DOCKER_BUILDKIT=0` in front of the `ktd` command,
-
-e.g: `DOCKER_BUILDKIT=0 ktd up`.
-
-This applies to all `Dockerfile`s where an insecure registry is specified.
-
 ```sh
 # Deploy the project to a Kubernetes namespace
 kubetools deploy my-namespace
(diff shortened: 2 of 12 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ### Unreleased
 
+# v20.0.0
```
gchazot: What about `14` ??

kubetools/dev/backends/docker_compose/docker_util.py
```diff
@@ -137,7 +138,12 @@ def get_containers_status(
             env = compose_project.replace(docker_name, '')
 
         # Where the name is compose-name_container_N, get container
-        name = container.name.split('_')[1]
+        print(container.name)
+        print(container.labels.get('kubetools.project.env'))
+        if int(__version__.split('.')[0]) < 20:
+            name = container.name.split('_')[1]
+        else:
+            name = container.name.split('-')[1]
```
gchazot: Seems like a useless check, this code will only ever be used with this version of the tool, and `__version__` is AFAIK the version of the tool, nothing else

gchazot (commented):
I doubt this is going to work as-is. It would be very lucky that all of the features of `docker-compose` that `kubetools` uses are fully backwards-compatible in `docker compose`.

CHANGELOG.md
fmcwalters-edited: that would make sense 👍 I only put this here as a placeholder so I could test the new version easily. Will change when PR is ready

kubetools/dev/backends/docker_compose/docker_util.py
fmcwalters-edited: > I doubt this is going to work as-is. It would be very lucky that all of the features of docker-compose that kubetools uses are fully backwards-compatible in docker compose.

this is why I have stuck this in draft, it still needs a full test across a range of OSes, archs and with as many of our services as possible.

fmcwalters-edited:
> I doubt this is going to work as-is. It would be very lucky that all of the features of docker-compose that kubetools uses are fully backwards-compatible in docker compose.

This is why it's only in draft, it still needs a thorough test across all dev machine OSes, architectures and with as many packages/services as possible.

kubetools/dev/backends/docker_compose/docker_util.py
fmcwalters-edited: Also still has print statements! I will remove those alongside the check

fmcwalters-edited:
I'm also seeing 
```
WARN[0000] network default: network.external.name is deprecated. Please set network.name with external: true 
```
from running `ktd destroy` - whilst this does not break anything _yet_, I feel like at this stage it is worth fixing.

kubetools/dev/backends/docker_compose/docker_util.py
fmcwalters-edited: [a439c4c](https://github.com/EDITD/kubetools/pull/163/commits/a439c4cf81449d620986cec9e67d3119f08ba093)

fmcwalters-edited:
force push because the previous code was weird - it was just to deal with the cases where there was more than one middle name

fmcwalters-edited:
we can test this locally using something like `echo -n "cython <3" > /tmp/constraint.txt && PIP_CONSTRAINT=/tmp/constraint.txt pip install -e ~/apps/kubetools` to build and install from this branch. Then ensure you're using v14.0.0 when you run `ktd`.

fmcwalters-edited:
I have put this up for a `dev` release - it has been tested to work (with caveats - see `README`) on Mac Intel i5 and Mac M3.
However - we need to test this on Linux systems and in particular on build servers.

README.md
```diff
@@ -80,15 +80,27 @@ cronjobs:
...
 ## Installing
 
+**NOTE**: before upgrading to version 14.0 or above, you _must_ run `ktd destroy` for all existing 
+local Kubetools projects.
+There is also a bug on the latest version of Docker Desktop that prevents the new Buildkit working with 
+insecure registries. This has been patched but the patch has yet to make its way to the latest release
+of Docker Desktop - [issue here](https://github.com/docker/buildx/issues/2030). 
+There are a couple of workarounds:
+  *  Downgrade to Docker Desktop v4.26.0 or below and append `http://` to each of your `insecure_registries`' URLs
+  *  Alternatively, use the legacy builder by appending `DOCKER_BUILDKIT=0` in front of every `ktd` command.
```
gchazot:
```suggestion
There are a couple of workarounds:
  * Migrate to secure registries (over HTTPS)
  * Downgrade to Docker Desktop v4.26.0 or below and prefix with `http://`, each of your `insecure_registries`' URLs.
  *  Alternatively, use the legacy builder by setting the environment variable `DOCKER_BUILDKIT=0` for `ktd` commands.
```

README.md
```diff
@@ -80,15 +80,27 @@ cronjobs:
...
 
+```sh
 # Deploy the project to a Kubernetes namespace
 kubetools deploy my-namespace
 ```
 
 ## Installing
 
+**NOTE**: before upgrading to version 14.0 or above, you _must_ run `ktd destroy` for all existing 
+local Kubetools projects.
```
gchazot: I believe this note belongs in CHANGELOG.md because it's a mostly a matter for people upgrading from 13 to 14, and there shouldn't be anything to do special for any new project.

README.md
```diff
@@ -80,15 +80,27 @@ cronjobs:
...
+There is also a bug on the latest version of Docker Desktop that prevents the new Buildkit working with
```
gchazot: For the "bug" aspect of it:
* it looks like it's a bug of `buildx` or `moby`, but not "Docker Desktop" which is a distribution of Docker. 
* it would be good to point at a/the ticket that is still open, or fixed, not one that's closed as "not planned"
* this section should be moved out of this section which is about "installing" the tool. It should be moved to a troubleshooting section.
* please add the latest version number of Docker( or moby or buildx) at the time of the PR, and that you have confirmed (tested) to have the bug.

kubetools/dev/backends/docker_compose/docker_util.py
```diff
@@ -136,8 +136,14 @@ def get_containers_status(
...
-        # Where the name is compose-name_container_N, get container
-        name = container.name.split('_')[1]
+        # if the container was made in v1, it will be composename_container_N
+        converted_name = container.name.replace('_', '-')
+
+        # we need to keep the middle part of the name
+        # for example if we have a container called `composename-container-1-N`
+        # we want to keep `container-1`
+        middle_names = converted_name.split('-')[1:-1]
+        name = '-'.join(middle_name for middle_name in middle_names)
```
gchazot: Can you please add tests for this? The main goal being to document the behaviour of this logic and its outcome when applied in each of the possible situations.

Note that the added test "`complex_named_app`" doesn't cover `docker-compose` code. These tests only cover `kubernetes` config generation.

fmcwalters-edited:
force push only to clean up some mistakes I made with linting and syntax 🤦‍♂️

kubetools/dev/backends/docker_compose/docker_util.py
fmcwalters-edited: I've factored this bit of the code out and added tests for different scenarios, plus added test coverage for the `dockerise_label()` function on which this is heavily reliant:
 [b9cc5a4](https://github.com/EDITD/kubetools/pull/163/commits/b9cc5a4cefc1758b48f11241471835c62c3ce99f)

README.md
fmcwalters-edited: I have added a better explanation of the bug itself, with a link to the relevant issue in `buildx`, plus a more detailed explanation of what happens if the user is affected by the bug and more detail in the workaround.
I've also cleaned up the CHANGELOG to remove information that shouldn't be there:
[7a8b920](https://github.com/EDITD/kubetools/pull/163/commits/7a8b920979214f3b3731fce5706bfd3f56f72628)

README.md
fmcwalters-edited: Have moved this within this commit:
[7a8b920](https://github.com/EDITD/kubetools/pull/163/commits/7a8b920979214f3b3731fce5706bfd3f56f72628)

kubetools/dev/backends/docker_compose/config.py
```diff
@@ -248,9 +248,8 @@ def create_compose_config(kubetools_config):
     if dev_network:
         compose_config['networks'] = {
             'default': {
-                'external': {
-                    'name': 'dev',
-                },
+                'name': 'dev',
```
gchazot, line 251: I'm wondering if we should take this opportunity to rename this "default" network used by `ktd`.

`default` here in the config is from the point of view of the project/container. This declares the default network for the project to be "external" (i.e. pre-existing outside this `compose` project) and named `dev`.

I think we should replace `dev` by something that clearly identifies it as used by `ktd`. Maybe `ktd_dev` could be an option.


Also, from a few tests I've run, it seems like other "environments" as created with `--env` option (or `test` in the case of `ktd test`), end up with a "temporary" network name (i.e. it's not "external" and IS deleted by the destroy operation) which looks like `<appname>_<env>_default`. I think it would be cool if that network name ended up being similarly, something like `ktd_<env>`.
This aspect probably requires a lot of work to implement, However it could be the base for a feature allowing to create "multi-app" environments, in a similar way as we can today do with the "default" environment.

README.md
```diff
@@ -137,6 +139,23 @@ minikube delete
 ...
 ```
 
+
+## Troubleshooting
+There is a bug in `buildx` which is present in Docker Engine v25.0 and up, which is yet to be patched and causes `insecure-registries` to be ignored - [issue here](https://github.com/docker/buildx/issues/2226).
+Versions of Docker Desktop which use versions of Docker Engine lower than 25.0 are unaffected.
+
+If you try and run `ktd up` with reference to an unsecure registry, e.g. `http://docker-registry.example.net` and are affected by this bug, you will get an error message that is
+similar to the following:
```
gchazot, line 152:
```suggestion
If you try and run `ktd up` with reference to an unsecure registry, e.g. `http://docker-registry.example.net` and are affected by this bug, you will get an error message that is similar to the following:
```

tests/test_docker_util.py
```diff
@@ -0,0 +1,37 @@
...
+from kubetools.dev.backends.docker_compose.docker_util import _get_container_name_from_full_name
+
+
+def generate_long_names(number_of_separators, separator):
+    return separator.join([
+        ch for ch in ascii_lowercase[:number_of_separators+1]])
+
+
+class TestDockerComposeNameConversion(TestCase):
+    # we need to check this works for dashes and underscores
```
gchazot, line 14:
```suggestion
```

tests/test_docker_util.py
```diff
@@ -0,0 +1,37 @@
+from string import ascii_lowercase
+from unittest import TestCase
+
+from kubetools.dev.backends.docker_compose.config import dockerise_label
+from kubetools.dev.backends.docker_compose.docker_util import _get_container_name_from_full_name
+
+
+def generate_long_names(number_of_separators, separator):
```
gchazot, line 8: See comment below, this function probably will disappear. But if it had to stay, then it needs to be renamed because it's confusing::
* It is generating ONE name, not  (several) `_names`
* Why do they have to be "long" ? Is it important for the tests? What would not be tested if the names were "short" ? Note that this aspect of making tests more understandable does not belong in this function's name only, but in how/why the result of it is used in tests.

tests/test_docker_util.py
```diff
@@ -0,0 +1,37 @@
...
+        ch for ch in ascii_lowercase[:number_of_separators+1]])
+
+
+class TestDockerComposeNameConversion(TestCase):
+    # we need to check this works for dashes and underscores
+    def test_dockerise_label_dashes(self):
+        for i in range(5):
+            long_name = generate_long_names(i, '-')
+            dockerised_name = dockerise_label(long_name)
+            self.assertEqual(dockerised_name, ascii_lowercase[:i+1])
```
gchazot, line 19: I'm not quite sure but I have a feeling that we're using `ascii_lowercase` here as a "long string", which "could look like an app name". If that the case, then in my opinion it's a bad idea. It's a case of "wrong abstraction". Instead here we want to use names that could really be what they stand for, so that it's obvious to the reader what we're talking about. Put them in a list and iterate over them. If you're worried about repeating that list too much, extract it in a constant and reuse it in each place.

kubetools/dev/backends/docker_compose/docker_util.py
```diff
@@ -192,6 +191,17 @@ def get_containers_status(
     return env_to_containers.get(kubetools_config['env'], {})
 
 
+def _get_container_name_from_full_name(full_name):
+    # if the container was made in v1, it will be composename_container_N
+    converted_name = full_name.replace('_', '-')
```
gchazot, line 196: Just to check, are we trying to create logic that is backwards compatible with compose V2?

I think that, in line with the change to README and with the major version change, we don´t need to support any backwards compatible logic here.

tests/test_docker_util.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def test_dockerise_label_underscores(self):
+        for i in range(5):
+            long_name = generate_long_names(i, '_')
+            dockerised_name = dockerise_label(long_name)
+            self.assertEqual(dockerised_name, ascii_lowercase[:i+1])
+
+    def test_container_name_from_full_name_dashes(self):
```
gchazot, line 27: These test names don't explain the scenario that we're testing here.
What are the cases where this logic might be relevant?

setup.py
```diff
@@ -59,8 +59,6 @@ def get_readme_content():
             # To support CronJob api versions 'batch/v1beta1' & 'batch/v1'
             'kubernetes>=21.7.0,<25.0.0',
             'tabulate<1',
-            # compose v2 has broken container naming
-            'docker-compose<2',
```
DilaraOflaz, line 63: Instead of removing `'docker-compose<2'`, should we add `'docker-compose,'` without specifying a version ?

ryan109: Docker-compose v2 isn't a python package anymore, I think it's go? Or it's the part of the docker package? I need to refresh my memory on that...

ryan109: https://docs.docker.com/compose/migrate/

fmcwalters-edited: `compose` is now a subcommand of `docker` itself, and consequently is a Go package that comes with every new Docker installation. There isn't _technically_ `docker-compose` v2, there's `docker-compose` which is a Python package and is now no longer maintained, and `compose` which is a feature of `docker`.

gchazot (commented):
Just realised I had all those comments pending from months ago.

PR #164, opened 2024-03-11
Use python docker compose

gchazot:
## Purpose of PR
When `ktd` is invoked, it used to call `docker-compose` command from the current PATH. This causes conflict with newer versions of docker which install their own version of `docker-compose` in the PATH. The version from Docker is using the "Compose V2" standard which is not yet supported by `kubetools` (WIP: #163).

This change makes sure that we call the "python" version of `docker-compose`. By using the current "executable", i.e. the current interpreter which is running `ktd`, and calling `compose` as a Python module, we maximise the chances that the version of docker-compose used is the Python one that was installed as a dependency of `kubetools`.

This creates the opportunity to install `kubetools` in its own separate virtualenv and run it by linking `kubetools` and `ktd` executables into the PATH. In a similar fashion, this should allow the support of `pipx` as an installation method.

### DIFF
--- a/.github/workflows/publish_pypi.yml
+++ b/.github/workflows/publish_pypi.yml
@@ -6,9 +6,9 @@ jobs:
   publish-to-pypi:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: Set up Python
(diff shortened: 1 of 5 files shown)


DilaraOflaz (approved):

gchazot:
Thanks @DilaraOflaz

PR #165, opened 2024-07-02
PFM-1685 / Add annotations option to resources defined in kubetools.yaml file

DilaraOflaz:
## Purpose of PR

---

I have tried on the following resources with and without `annotations`
```
name: telegraf-test

deployments:
  telegraf-test:
    annotations:
       app.kubernetes.io/name: "telegraf-test"
    containers:
      telegraf-test:
        image: test-image
        

dependencies:
  mariadb:
    annotations:
         app.kubernetes.io/name: "telegraf-test"
    containers:
      mariadb-container:
        image: test-images

cronjobs:
  my-cronjob:
    annotations:
       app.kubernetes.io/name: "telegraf-test"
    schedule: "*/1 * * * *"
    concurrency_policy: "Replace"
    containers:
      hello:
        image: test-images
        command: [/bin/sh, -c, date; echo Hello from the Kubernetes cluster]

```

## Related links

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,10 @@ tests:
 
 deployments:
   my-app-webserver:
+    annotations:
+      imageregistry: "https://hub.docker.com/"
+    labels:
+      app.kubernetes.io/name: my-app-webserver
     serviceAccountName: webserver
     secrets:
       secret-volume:
@@ -59,8 +63,6 @@ deployments:
           - 80
         dev:
           command: [./manage.py, runserver, '0.0.0.0:80']
-        annotations:
-          imageregistry: "https://hub.docker.com/"
 
 dependencies:
   mariadb:
--- a/README.md
+++ b/README.md
@@ -59,6 +59,8 @@ deployments:
           - 80
         dev:
           command: [./manage.py, runserver, '0.0.0.0:80']
+        annotations:
+          app.kubernetes.io/name: "django_app"
 
 dependencies:
   mariadb:
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ deployments:
         dev:
           command: [./manage.py, runserver, '0.0.0.0:80']
         annotations:
-          app.kubernetes.io/name: "django_app"
+          imageregistry: "https://hub.docker.com/"
 
 dependencies:
   mariadb:
--- a/kubetools/kubernetes/config/__init__.py
+++ b/kubetools/kubernetes/config/__init__.py
@@ -183,8 +183,7 @@ def generate_kubernetes_configs_for_project(
         )
         app_annotations = copy_and_update(base_annotations)
         for container_name, data in containers.items():
-            if data.get('annotations'):
-                app_annotations.update(data.get('annotations'))
+            app_annotations.update(data.get('annotations', {}))
 
         if container_ports:
             services.append(make_service_config(
@@ -230,8 +229,7 @@ def generate_kubernetes_configs_for_project(
             per_deployment_annotations.get(name),
         )
         for container_name, data in containers.items():
-            if data.get('annotations'):
-                app_annotations.update(data.get('annotations'))
+            app_annotations.update(data.get('annotations', {}))
 
         if container_ports:
             services.append(make_service_config(
@@ -333,8 +331,8 @@ def generate_kubernetes_configs_for_project(
 
         app_annotations = copy_and_update(base_annotations)
         for container_name, data in containers.items():
-            if data.get('annotations'):
-                app_annotations.update(data.get('annotations'))
+            app_annotations.update(data.get('annotations', {}))
+
         schedule = cronjob['schedule']
         concurrency_policy = cronjob['concurrency_policy']
         batch_api_version = cronjob.get('batch-api-version')  # May depend on target cluster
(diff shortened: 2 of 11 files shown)


kubetools/kubernetes/config/__init__.py
```diff
@@ -182,6 +182,9 @@ def generate_kubernetes_configs_for_project(
             default_registry=default_registry,
         )
         app_annotations = copy_and_update(base_annotations)
+        for container_name, data in containers.items():
+            if data.get('annotations'):
+                app_annotations.update(data.get('annotations'))
```
gchazot: This is equivalent to (x3)
```suggestion
            app_annotations.update(data.get('annotations', {}))
```

gchazot (changes requested):
Please add tests to show it works as expected

kubetools/kubernetes/config/__init__.py
DilaraOflaz: https://github.com/EDITD/kubetools/pull/165/commits/02c60076d29e2b4daaa303248cda84b8d7c7dd0b

DilaraOflaz:
Here are the tests @gchazot 
https://github.com/EDITD/kubetools/pull/165/commits/b137d4fe92fba19f7f48e009db4d5e980d9b3878

kubetools/kubernetes/config/container.py
```diff
@@ -128,6 +128,9 @@ def make_container_config(
                 'readonly': True,
             })
 
+    # Remove annotations from container data
+    if 'annotations' in container:
+        container.pop('annotations')
```
DilaraOflaz: I had to add this otherwise the annotations was also added to the container definition, see the last 3 lines of the example  below:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  annotations: {
    app.kubernetes.io/managed-by: kubetools,
    app.kubernetes.io/name: generic-app-with-annotations
  }
  labels: {
    kubetools/name: generic-app-with-annotations,
    kubetools/project_name: generic-app,
    kubetools/role: app
  }
  name: generic-app-with-annotations
spec:
  replicas: 1
  revisionHistoryLimit: 5
  selector:
    matchLabels: {
      kubetools/name: generic-app-with-annotations,
      kubetools/project_name: generic-app,
      kubetools/role: app
    }
  template:
    metadata:
      labels: {
        kubetools/name: generic-app-with-annotations,
        kubetools/project_name: generic-app,
        kubetools/role: app
      }
    spec:
      containers:
      - command: [generic-command]
        containerContext: generic-context
        env:
        - {name: KUBE, value: 'true'}
        image: generic-image
        imagePullPolicy: Always
        livenessProbe:
          httpGet: {path: /ping, port: 80}
          timeoutSeconds: 5
        name: webserver
        readinessProbe:
          httpGet: {path: /ping, port: 80}
          timeoutSeconds: 5
        annotations: {
          app.kubernetes.io/name: generic-app-with-annotations
         }
```

README.md
```diff
@@ -59,6 +59,8 @@ deployments:
           - 80
         dev:
           command: [./manage.py, runserver, '0.0.0.0:80']
+        annotations:
+          app.kubernetes.io/name: "django_app"
```
gchazot: What's the goal of this by the way?
I think there's some confusion here because this specific tag is supposed to be a _label_, not an _annotation_  (see [here](https://kubernetes.io/docs/reference/labels-annotations-taints/#app-kubernetes-io-name) and [here](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/))

More generally, I think that we should add the option to add both: labels and annotations

kubetools/kubernetes/config/__init__.py
```diff
@@ -182,6 +182,8 @@ def generate_kubernetes_configs_for_project(
             default_registry=default_registry,
         )
         app_annotations = copy_and_update(base_annotations)
+        for container_name, data in containers.items():
+            app_annotations.update(data.get('annotations', {}))
```
gchazot: I am quite concerned here with the structure.

We are raising the "annotations" from the container level to the "app" level (which I think means the "deployment" / "job" / ... level) . For example, what happens if:
* we want to add annotations at the container level?
* we define the same annotation with different values on 2 containers of the same "app"?

I think the root problem here is that the annotations should not be taken from the container level but from the "app" level, if the intent is to label the "app"

kubetools/kubernetes/config/container.py
gchazot: This is the consequence of the problem I'm referring to in another comment

gchazot (commented):
I'm concerned by the structure added here, see comments

README.md
DilaraOflaz: My bad, I copied a bad example
Here https://github.com/EDITD/kubetools/pull/165/commits/b9d828cca05e2b3ec5b95afedc2874c09397495f

kubetools/kubernetes/config/container.py
DilaraOflaz: It has been removed in this commit https://github.com/EDITD/kubetools/pull/165/commits/12c0b0863d2a93f7544afcaf3288c7c82db5d899

kubetools/kubernetes/config/__init__.py
DilaraOflaz: https://github.com/EDITD/kubetools/pull/165/commits/12c0b0863d2a93f7544afcaf3288c7c82db5d899

gchazot (approved):
LGTM

PR #168, opened 2025-12-12
Update Docker Compose integration to use v2 CLI and get container name

Luladjiev:
from labels

This pull request introduces several important updates in preparation for version 14.0.0, including support for Docker Compose v2, removal of Python 3.7 support, and modernization of some dependencies and internal APIs. The changes also update the GitHub Actions test matrix and streamline Docker Compose integration.

**Key changes:**

**Docker Compose v2 support and related improvements:**
* Updated Docker Compose integration to use the `docker compose` CLI (instead of the Python module or `docker-compose`), enabling support for Docker Compose v2. This includes changing how compose processes are run and how container service names are extracted, ensuring compatibility across Compose versions. [[1]](diffhunk://#diff-9e7acd433c2a68cab42ea6663574b6d9ea712c5a3a456cde473a501121d42aedL207-R205) [[2]](diffhunk://#diff-9e7acd433c2a68cab42ea6663574b6d9ea712c5a3a456cde473a501121d42aedL141-R140) [[3]](diffhunk://#diff-0ddc34c607283bbd8573fb06c612c0d74ad68ac252d3a427ea6ec2cbe86d6e53L251-R252)
* Removed the `docker-compose<2` dependency and related comments from `setup.py`, reflecting the move to Compose v2 and the CLI approach.

**Python version and dependency updates:**
* Dropped support for Python 3.7 and updated the GitHub Actions test matrix to test only supported Python versions (3.8+), and updated the tested Ubuntu versions. [[1]](diffhunk://#diff-c2e9f7c1bfaf65da79bbd29ed3ce389b2acc8b4099a257914a7292a66e6b9bc9L9-R10) [[2]](diffhunk://#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR5-R8)

**Dependency and import modernization:**
* Replaced deprecated `pkg_resources` usage with `importlib.metadata` for version retrieval in `__init__.py`.
* Switched from `pkg_resources.parse_version` to `packaging.version.parse` for version parsing in `config.py`.

**Changelog:**
* Added a new section for v14.0.0 in `CHANGELOG.md` documenting Docker Compose v2 support and Python 3.7 deprecation.

### DIFF
--- a/kubetools/config.py
+++ b/kubetools/config.py
@@ -7,7 +7,7 @@ from os import getcwd, path
 
 import yaml
 
-from pkg_resources import parse_version
+from packaging.version import parse as parse_version
 
 from . import __version__
 from .exceptions import KubeConfigError
--- a/setup.py
+++ b/setup.py
@@ -59,8 +59,6 @@ if __name__ == '__main__':
             # To support CronJob api versions 'batch/v1beta1' & 'batch/v1'
             'kubernetes>=21.7.0,<25.0.0',
             'tabulate<1',
-            # compose v2 has broken container naming
-            'docker-compose<2',
         ),
         extras_require={
             'dev': (
--- a/setup.py
+++ b/setup.py
@@ -55,7 +55,6 @@ if __name__ == '__main__':
             'pyyaml>=3,<6',
             'requests>=2,<2.29.0',  # https://github.com/docker/docker-py/issues/3113
             'pyretry',
-            'setuptools',
             # To support CronJob api versions 'batch/v1beta1' & 'batch/v1'
             'kubernetes>=21.7.0,<25.0.0',
             'tabulate<1',
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -6,7 +6,7 @@ jobs:
   tests:
     strategy:
       matrix:
-        os: [ubuntu-20.04, ubuntu-latest]
+        os: [ubuntu-22.04, ubuntu-latest]
         python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
     runs-on: ${{ matrix.os }}
     steps:
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -7,11 +7,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-20.04, ubuntu-latest]
-        python-version: ['3.6', '3.7', '3.8', '3.9', '3.10', '3.11', '3.12']
-        exclude:
-          # Python 3.6 is not available in GitHub Actions Ubuntu 22.04
-          - os: ubuntu-latest
-            python-version: '3.6'
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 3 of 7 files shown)


Luladjiev:
@EDITD/platform This PR allows developers using Podman to use Kubetools for development. It was tested by some of the developers already and we didn't stumble upon any issues.

kubetools/config.py
```diff
@@ -7,7 +7,7 @@
 
 import yaml
 
-from pkg_resources import parse_version
+from packaging.version import parse as parse_version
```
seangarner, line 10: Where is packaging coming from?  It's not in setup.py so I guess it's a dep of one of our other deps which over time might not be a safe assumption.

setup.py
seangarner, line 51: If we're dropping python 3.7 support then shouldn't this be updated?

fmcwalters-edited:
I started looking into this ages ago and we found problems
https://github.com/EDITD/kubetools/pull/163

@ashirza also has been doing work trying to make this change and has run into similar problems

fmcwalters-edited:
https://edited.atlassian.net/browse/PFM-827

this will break http with `ktd` for developers. Our docker-registry uses http.

Luladjiev:
@fmcwalters-edited As I mentioned in my previous comment, we’ve been using it for the past couple of months with Podman without any issues. Also if I remember correctly @svedbg tested it on his machine with docker and it worked. Is it possible someone from platform to test with Docker Desktop if it works or not, as we already moved to Podman?

kubetools/config.py
Luladjiev: https://github.com/EDITD/kubetools/pull/169

setup.py
Luladjiev: https://github.com/EDITD/kubetools/pull/169

ashirza:
moving my comment over from the other branch:

I'm very happy someone else is interested in modernising kubetools 😄 would love to work together on this if we're able to get past the insecure registry blocker/other blockers. I've been using my own custom version locally for a little while now, that I've shared with other members of the team.

I plan to try to test this out with podman - it would be great if we don't run into the same issue that we do with docker/rancher desktop with insecure registries & the underscore/hyphen breaking changes. I do believe that some devs are still using docker desktop, so we should keep this in mind.
I'm not sure what work would need to done in CI (jenkins) here if we find that we're able to proceed as is.

As @fmcwalters-edited mentioned, I've also been doing some work here on this too 😅 most of these were fairly similar (dropping support for 3.6, 3.7), I tried to minimise the changes I made while still modernising kubetools over a few branches (that I haven't pushed yet):
- the use of [python-on-whales](https://github.com/gabrieldemarmiesse/python-on-whales), dropping the docker & docker compose python SDKs.
  - As you probably know, the python docker compose SDK won't work with compose v2. The docker SDK hasn't seen a release for 2 years either (although I did see a merge ~4 months ago). Something to consider as a team.
- Parallel builds, stops etc.
- the use of compose v2 `--compatibility` mode, which lets us avoid having to run into the underscore/hyphen issue. although it seems like you've got a change here that accounts for this.
- bumping `pyyaml`, so we don't need to pin `cython` every single time we try to install kubetools
- flake8 -> ruff
- py3.13 support

